PR #33, opened 2010-10-25
Fixed a problem that wouldn't let me run mrjob

jretz:
When trying to run one of the example jobs, I see the stack trace reproduced below. This change fixes that along with a trivial wording mistake in a comment.

```
Traceback (most recent call last):
  File "dog.py", line 1, in <module>
    from mrjob.job import MRJob
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/mrjob/job.py", line 599
    '--step-num', dest='step_num', type='int', default=0, default=False,
SyntaxError: keyword argument repeated
```

### DIFF
--- a/mrjob/job.py
+++ b/mrjob/job.py
@@ -366,7 +366,7 @@ class MRJob(object):
             return LocalMRJobRunner(**self.local_job_runner_kwargs())
 
     def run_job(self):
-        """Run the all steps of the job, logging errors (and debugging output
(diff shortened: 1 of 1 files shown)


Sana1221:
python2: can't open file 'dog.py': [Errno 2] No such file or directory

Sana1221:
What do this time 👆

PR #447, opened 2012-05-08
Integrate typedbytes into mrjob

### LINKED ISSUE #430: Add TypedBytes support to mrjob
One of the major differences between mrjob and dumbo is that dumbo uses the typedbytes interface to communicate with the hadoop streaming runtime.  We are working on adding typedbytes to mrjob for this purpose.  

There was some preliminary discussion of this on the dumbo docs issue #11.

The current state of the patch is joshi-prahlad@f62e6f6458270304b62a09579ea27f36f8dc9efc

dgleich:
Here is a patch to add support for typedbytes to mrjob.

To enable a class to use typedbytes instead of the textual interface, you just need to specify

class MyMRJob(MRJob):
    STREAMING_INTERFACE = MRJob.STREAMING_INTERFACE_TYPED_BYTES

And then the interface to hadoop streaming will use typedbytes instead of line based formats.

This is compatible with dumbo, and adds typedbytes as a requirement to MRJob.

This fixes #430.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,7 +1,57 @@
-v0.3.2, 2012-??-?? -- ???
- * mrboss no longer spells stderr 'stsderr'
- * 'Testing with mrjob' section in docs (includes #321)
- * ami_version option to allow switching between EMR AMIs (#306)
+v0.3.3.2, 2012-04-10 -- It's a race [condition]!
+ * Option parsing no longer dies when -- is used as an argument (#435)
+ * Fixed race condition where two jobs can join same job flow thinking it is
+   idle, delaying one of the jobs (#438)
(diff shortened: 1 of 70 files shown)


davidmarin:
So, I was pretty much on hiatus as a mrjob maintainer for a year, and I'm back.

Are you still interested in helping out with this change?

Honestly, first priority is to get the 0.4.3 release out, and second priority is to get mrjob working smoothly with the new AMIs, but after that I can get to reworking the way we handle protocols.

davidmarin:
Closing this pull request because there's no intention to merge it as-is, and it looks terrible to have a 2-year-old pull request. I haven't given up on this issue though!

PR #885, opened 2014-03-10
Writeable filesystem

icio:
Defines `{Composite,Local,S3,Hadoop}Filesystem.write`.


ikreymer:
Are there currently any plans to integrate any version of this into a future release? The `write` and `copy_from_local` would be super useful, want to avoid having to reimplement. Thanks

icio:
@ikreymer we're currently using https://github.com/duedil-ltd/pyfilesystem/pull/2 instead of this but it requires the WebHDFS API available in Hadoop 2.

PR #911, opened 2014-05-13
Added ability to specify the step action on failure via the emr_action_o...

countrodrigo:
...n_failure option

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -478,6 +478,15 @@ Choosing/creating a job flow to join
     flow every 30 seconds until this many minutes have passed, then start a new
     job flow instead of joining one.
 
+.. mrjob-opt::
+    :config: emr_action_on_failure
+    :switch: --emr-action-on-failure
+    :type: :ref:`string <data-type-string>`
+    :set: emr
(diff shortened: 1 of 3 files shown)


davidmarin:
Thanks @countrodrigo. Looking into this! Can you tell me more about your use case?

countrodrigo:
Currently the default behavior is to cancel all queued up steps and wait for new ones to be submitted. We run many independent steps that we queue up on the cluster, and which should continue even if a prior step failed, thus we prefer to specify CONTINUE instead.

davidmarin:
Thanks! That totally makes sense. Will try to slip it into the 0.4.3 release.

By the way, do you have any idea what the difference between `TERMINATE_JOB_FLOW` and `TERMINATE_CLUSTER` is? I haven't been able to find the relevant documentation.

davidmarin:
Never mind found it [here](http://awsdocs.s3.amazonaws.com/ElasticMapReduce/latest/emr-api.pdf): "`TERMINATE_JOB_FLOW` is deprecated, use `TERMINATE_CLUSTER` instead."

davidmarin:
Incorporated this into pull request #976, now awaiting review.

PR #942, opened 2014-08-22
Virtualenv

danjamker:
Change -e to -d for detecting if the virtualenv is already created


msabramo:
Bump

davidmarin:
Sorry, your whitespace is messy and I'm not convinced this adds any value. If the path exists but isn't a directory, `virtualenv` is doomed anyway.

Issue #713, opened 2013-08-28
drop Python 2.5 support


davidmarin:
We should drop support for Python 2.5 in preparation for supporting Python 3 (#217).

Mostly this means:
 * remove `from __future__ import with_statement`
 * use `except FooError as bar` rather than `except FooError, bar`
 * update setup.py
 * update docs

davidmarin:
Fixed by #945.

Issue #217, opened 2011-09-12
Python 3.3 compatibility (except EMR)


irskep:
The goal of this ticket is a full port of mrjob that works for everything except EMR. This includes tests.

Roadmap:

* v0.4.2: Last release to support 2.5
* v0.4.4: Experimental support for Python 3, except EMR
* v0.5.0: Full support for Python 3.

This does not include EMR support (#989) or updating docs and examples (#994).

bonitao:
Hi @irskep,

I have finished the work started by @edschofield. I have forked his repo, and you can see the remaining changes for python 3 compatibility here: https://github.com/bonitao/mrjob/tree/py3.

The changes are all sort of straightforward, and the biggest headache is python 3 improved discipline on encodings. You can pull my repository, but since there may be too many changes, I will be happy to send you piecemeal, logically contained, commits if you prefer. 

All tests pass in python2 and python3, and I have ran a few jobs locally or inline. I have not tried it on aws yet though, but I will do it in the next few weeks.

irskep:
Thanks! I think @anusha-r is the current maintainer who will need to look at it.

davidmarin:
Realized `assertItemsEqual()` won't work on Python 3.

davidmarin:
Fixed by #1007

PR #945, opened 2014-09-06
Towards Python 2/3 compatibility (take 2)

edschofield:
Here is a set of patches that takes ``mrjob`` a step toward towards Py2/3 compatibility. Python 3 is not yet supported, but the ``diff`` here should be easy to review and take us in the right direction.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ mrjob: the Python MapReduce library
 
 .. image:: docs/logos/logo_medium.png
 
-mrjob is a Python 2.5+ package that helps you write and run Hadoop Streaming
+mrjob is a Python 2.6+ package that helps you write and run Hadoop Streaming
 jobs.
 
 `Stable version (v0.4.2) documentation <http://packages.python.org/mrjob/>`_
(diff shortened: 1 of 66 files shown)


msabramo:
This doesn't merge cleanly anymore. Hopefully the maintainers can try to merge it soon after the conflicts are fixed.

edschofield:
So far the maintainers haven't indicated that they are interested in it. If they do, I will update the patch.

davidmarin:
Thank you! I'd love to see `mrjob` working on Python 3, and this is a step in the right direction.

I'm not going to include this in the 0.4.3 release for fear of breaking things, but we are officially dropping support for Python 2.5 (see #713).

davidmarin:
The 0.4.3 release is all but done, so I'm working on merging your changes for the 0.4.4 release. I'm hoping to have experimental Python 3 support in 0.4.4 (see #217).

hayd:
@ bonitao has a branch on top of this which passes all the tests: https://github.com/bonitao/mrjob/tree/py3

davidmarin:
Neat, took a look. I'm actually most of the way through implementing a solution that doesn't require `six`.

davidmarin:
Rolled this into #1007.

PR #961, opened 2014-12-18
avoid import of mrjob library (and its dependencies) to version on setup.py

steenzout:
1. added a version module.
2. moved `mrjob.__version__` to `mrjob.version.__version__`
3. linked `mrjob.__version__` attribute to `mrjob.version.__version__` attribute.
4. added test case for `mrjob.version` module.

not really related but I also added 2 lines on .gitignore to ignore PyCharm .idea directory and .eggs directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.pyc
 *~
 *.swp
+.eggs
+.idea
 .project
 .pydevproject
 .ropeproject/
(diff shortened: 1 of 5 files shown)


msabramo:
:+1: importing a module in `setup.py` to get the version can cause installation problems. Doing `execfile` like here is much better.

davidmarin:
Thanks for the tip! We don't really need a `version` module though.

#973 uses `execfile()` and includes your commits, and your changes to `.gitignore`.

PR #963, opened 2014-12-28
Using non-local flag in inline runner when executing steps

### LINKED ISSUE #851: add_file_option() is broken in inline mode
Much of what is said here is repeated: 
https://groups.google.com/forum/#!topic/mrjob/iGzC3jRYoDY
 
In the page http://pythonhosted.org/mrjob/guides/writing-mrjobs.html#file-options it says that the value of the option is changed to be the path of the file. But this is not happening as documented for inline mode.

My MRJob subclass I have code to read the file(simplified):
```
def mapper_init(self):
   f = open(self.options.platform_file)
...
```

My test is sets up the MRJob subclass as follows:
```
   def test_testjob():
           my_job = MyJob(
               ['-r', 'inline', '--platform_file=myproject/tests/resources/configuration/platform.yml', '--no-conf',
               '-'])
...
```

I have set up the file options in my MRJob subclass:

```
class MyJob(MrJob):
    def configure_options(self):
        super(MyJob, self).configure_options()
        self.add_file_option('--platform_file')
......
```

My Project structure is as follows:
<ul> 
<li>myproject/</li>
<li>__init__.py</li>
<li>myjob.py</li>
  <ul>
  <li>tests/</li>
  <li>__init__.py</li>
  <li>my_test.py</li>
    <ul>
      <li>resources/configuration/</li>
        <ul>
          <li>platform.yml</li>
        </ul>
    </ul> 
  </ul> 
</ul>

But what ends up happening is that if I use the "inline" mode. The path substitution does not happen. This seems to be a bug. 

NOTES:
I see that there is a symlink to the platform file in a local dir:
(pylnrt)sid-mbp-2:pylnrt sidharth$ ls -lart /private/var/folders/s7/rxgcg85s3tv1ytgttl1t6_p40000gq/T/mytestjob.sidharth.20131231.170847.011298/job_local_dir/0/mapper/0/
total 16
lrwxr-xr-x  1 sidharth  staff  112 Dec 31 12:08 platform.yml -> /Users/sidharth/repo/myproject/tests/resources/configuration/platform.yml
but my self.options.platform_file is not set to the symlink.

Also note that the current working directory of the python vm when its executing the *_init() methods is the temporary directory. Which means that the file will need to be accessible from that absolute path.

anusha-r:
Fixes #851


bchess:
Ship it!

davidmarin:
Wow... this is really obscure. Had to go through the code to remind myself what this does. 

So, right, inline steps still run in their own temporary working directory, so they can reference the symlinks to files that were set up in the working directory rather than the real paths of those files.

Would be really worthwhile to have a unit test for this; otherwise some future maintainer is going to wonder why it says `local=True` in the method above and `local=False` in this method, and change it back.

If you can't afford a unit test, a comment will do.

Issue #805, opened 2013-10-31
Warn about use of deprecated bootstrap/setup options


davidmarin:
These are deprecated in v0.4.2, but we can wait until v0.4.3 before we start yelling at people.

PR #964, opened 2015-01-10
Adding deprecation warnings for setup/bootstrap* options

anusha-r:
This is a straightforward check against deprecated options mentioned in #805.
Not sure if there's a cleaner way to do this and to tackle differences in option names between config file and command line (setup_cmd VS --setup-cmd)

### DIFF
--- a/docs/guides/configs-all-runners.rst
+++ b/docs/guides/configs-all-runners.rst
@@ -359,6 +359,18 @@ Other
 
     .. versionadded:: 0.4.3
 
+.. mrjob-opt::
+    :config: export_job_name
+    :switch: --export-job-name
+    :type: boolean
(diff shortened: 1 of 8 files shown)


anusha-r:
@davidmarin : Can you take a look at this request and https://github.com/Yelp/mrjob/pull/963 when you get a chance?

bchess:
Ship it!

davidmarin:
Yup, this is the right way to do that.

Some minor nits about consistency:
 * some of the warnings say "Consider using bootstrap." and some say "Consider using bootstrap instead."
 * historically, mrjob's deprecation warnings have also told when a feature will be removed ("...and will be removed in v0.5.0")

PR #968, opened 2015-01-24
Wrapping contextmgr functions with try-catch

anusha-r:
I noticed that the order of tests caused failures: if test_job.StrictProtocolsTestCase ran before test_inline.MRJobCustomFileOptionsTestCase, the latter failed consistently. Turns out the StrictProtocolsTestCase has tests that simulate exceptions and some of the functions in mrjob/util.py were not fail-proof: their cleanup after yield did not happen in case of exceptions.
Wrapping those methods in a try-catch now. With this, the order of tests shouldn't be a problem anymore.


smallredbean:
Although try-catch solve the issue, it seems to me this is a hack-y solution.

davidmarin:
Good catch. From what I've found online, it looks like try/finally is exactly how you're supposed to do it; looks like I got it wrong. :)

PR #970, opened 2015-02-19
filter out any steps that have their kwarg set to None

bchess:
When explicitly setting step kwargs as None, i.e. mapper=None, reducer=None, it should not count as having an actual mapper or reducer.

Previously these end up counting as an "identity" process.


smallredbean:
Ship it

PR #973, opened 2015-03-09
execfile() in setup

davidmarin:
Use `execfile()` to get `mrjob.__version__` rather than importing it.

This is basically #961 without the `version` module.


msabramo:
LGTM although I didn't test and an unrelated changelog change seems to have snuck in. 

		+   * MRStep's constructor treats kwarg=None same as not setting it (#970)

Fix that and if it's tested I'd say it's good.

msabramo:
If you like the idea of managing the version with VCS tags, you could check out setuptools_scm. I just switched to it for pdbpp and I like it.

davidmarin:
The CHANGES.txt line is already in `master`; just diff weirdness.

davidmarin:
Thanks for the tip! We try to keep mrjob's dependencies to a minimum though.

PR #977, opened 2015-03-10
Eliminate `no_emr_api_params` option

davidmarin:
In #868, I asked @ashkop to remove the `no_emr_api_params` option "internally", which resulted in a warning every time we instantiated the EMR job runner:

```
Got unexpected keyword arguments: no_emr_api_params
```

Basically, I'd asked @ashkop to do something that was trickier than I thought. The `no_emr_api_params` option had to exist in the option parser, but we only wanted it so we could update `emr_api_params` in `MRJobLauncher.load_options()`, so we didn't want `EMRJobRunner` to think of it as an option, but that made it an unexpected keyword argument.

This patch fixes the warning, adds tests (it was previously untested), and the feature's documentation to the point that someone other than the implementor of this feature might figure out what it does and use it. :)

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -95,6 +95,47 @@ Job flow creation and configuration
     :mrjob-opt:`s3_endpoint`.
 
 .. mrjob-opt::
+    :config: emr_api_params
+    :switch: --emr-api-param, --no-emr-api-param
+    :type: :ref:`dict <data-type-plain-dict>`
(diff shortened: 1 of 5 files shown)


davidmarin:
Whoops, my bad, there were test cases. Going to combine the test classes.

davidmarin:
Okay, this is good to go.

zachm:
lgtm, good to see this error message go away!

PR #985, opened 2015-03-25
basics of the parts, with failing tests

squirrely:
This is the basics of what needs to be done. Known problems:
   - failing tests (configuration of emrjobrunner)
   - bad configuration...hardcoded :(

PR #986, opened 2015-03-31
Multipart upload to S3

davidmarin:
This is a heavily reworked version of #985 that enables multipart upload for large files (by default, larger than 100MB).

#985 had two (hard-coded options), a threshold to enable multipart uploading, and a minimum part size. I made these the same thing for simplicity.

Tests pass, but I still need to hand-test this change!

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -593,12 +593,20 @@ which is set to True if boto.Version >= '2.25.0'
    :set: emr
    :default: 100
 
-   Upload files to S3 in parts no bigger than this many megabytes. Default is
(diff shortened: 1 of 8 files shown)


squirrely:
This looks great to me. Thanks for taking this over! The tests are some brutal mocking well done.  :)

zachm:
@davidmarin lgtm. happy to merge once you're satisfied with your hand testing.

davidmarin:
Hand-tested! Of course I tested it with the smallest part size possible, and found out that Amazon's MB are actually MiB (`1024 * 1024`, not `1000 * 1000`). After that it passed.

Since this is a trivial code change to the version you approved, going to merge it myself. :)

PR #987, opened 2015-04-02
make all tests work without mrjob installed

### LINKED ISSUE #984: some tests can't find mrjob in tox
So, mrjob now has tox support #895, but some tests still fail with `ImportErrors`; basically, whatever script the test is running can't find mrjob. This is a known issue with mrjob: tests that invoke a subprocess can only find mrjob if it's *installed*.

I *think* we just need to tell tox to install mrjob (in the virtualenv it creates) before running the test.

davidmarin:
This fixes #984 and does away with the weird quirk that mrjob's tests could only pass if mrjob was installed.

Basically this does two things:
* re-enables bootstrapping mrjob in `local` mode (by deleting lines from `mrjob/sim.py`). This *does* change a default, but it isn't a breaking change, and this actually just re-aligns mrjob with its docs
* for the handful of tests where we don't want to bootstrap mrjob, or where we're running some other subprocess that needs mrjob (mock hadoop), we update `PYTHONPATH` so that our version of mrjob comes first.

I also fixed some related docs that happened to have been wrong for a while, now that local/inline mode simulate Hadoop.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -22,6 +22,8 @@ v0.4.3, 2014-03-?? -- SO many bugfixes
      * exclude hadoop source jars when looking for streaming jar (#861)
      * Fixed mkdir_on_hdfs for Hadoop version 2.x (#923)
      * Fixed hadoop_bin on Windows (#843)
+   * Local
+     * bootstrap mrjob by default (#984)
    * Inline
      * fix for add_file_option() (#851)
  * Use pytest to run tests (#898)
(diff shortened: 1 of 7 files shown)


davidmarin:
Can confirm that `tox` runs correctly now. Even installed `pypy` so I could test that.

davidmarin:
Since this is basically just a test/docs change (the only library code change was deleting code that sets a default), merging it myself.

PR #988, opened 2015-04-07
Inline fix

bchess:
instantiate the mrjob class inside the new working directory, because some mrjobs load files inside load_options()


davidmarin:
Thanks! Cherry-picked your change into `master`.

PR #995, opened 2015-04-08
Update EC2 dicts with more recent instance types

cmatthias:
I noticed that jobflow pooling wasn't working all that well for me, because I was using m3 and c3 instances, neither of which had ECU/memory stats in the dictionaries in aws.py. 

I took these values from https://aws.amazon.com/ec2/pricing/. The only questionable entries are the t2.* instances which have a "variable" number of ECUs according to amazon; I set these at 1 ECU each for the purposes of jobflow pooling, which is probably the most conservative thing to do.


davidmarin:
Thanks! Just in time for v0.4.3

davidmarin:
If you look at http://aws.amazon.com/ec2/instance-types/, the t2 instance types are "burstable", and receive a certain number of credits per hour (a credit is worth 1 CPU for 1 minute).

Since map-reduce usage is more-or-less continuous, I'm going to rate these instances by the actual amount of CPU they offer in an hour. For example, a `t2.micro` has 6 credits or 0.1 CPU/hour.

Thanks for pointing out this edge case!

cmatthias:
Oh cool, I completely missed that explanation on their site. That makes complete sense. Thanks for the quick merge!

davidmarin:
I do what I can!

PR #1004, opened 2015-04-21
work with new IAM requirements

### LINKED ISSUE #999: EMR requires jobflow/service roles for new accounts
This may make mrjob not work out-of-the-box for new users, specifically those with accounts created after April 6, 2015.

See [this Google Groups thread](https://groups.google.com/forum/#!topic/mrjob/h7-1UYB7O20). The underlying change in EMR is described [here](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-iam-roles-creatingroles.html).

New mrjob users shouldn't have to worry about this nonsense at all; mrjob should create the default roles for them.

davidmarin:
This fixes #999, which prevents mrjob from working for new AWS accounts. It's similar to automatic bucket creation; mrjob only creates IAM objects as needed, and once it's created them, they can be reused.

Also added an `iam_service_role` option and renamed the `iam_job_flow_role` option to `iam_instance_profile` because it's not just out-of-date, it's actually incorrect (IAM instance profiles are different from IAM roles). `iam_job_flow_role` still works; it's just deprecated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,6 @@ Setting up EMR on Amazon
 ------------------------
 
 * create an `Amazon Web Services account <http://aws.amazon.com/>`_
-* sign up for `Elastic MapReduce <http://aws.amazon.com/elasticmapreduce/>`__
 * Get your access and secret keys (click "Security Credentials" on
   `your account page <http://aws.amazon.com/account/>`_)
 * Set the environment variables ``$AWS_ACCESS_KEY_ID`` and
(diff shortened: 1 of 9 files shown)


zachm:
I am super okay with this change. Ship It!

davidmarin:
Travis appears to be stalled, but hand-tested on two different accounts, and ran the tests in `tox`. So, okay with it.

Issue #997, opened 2015-04-12
don't use unittest2 in Python 2.7


davidmarin:
`unittest2` is basically a shim to get Python 2.7 testing features in Python 2.6.

For the list of the (inconsequential, for us) differences between `unittest2` and `unittest` on Python 2.7, see https://pypi.python.org/pypi/unittest2.

davidmarin:
This should be fixed as soon as I merge https://github.com/davidmarin/mrjob/tree/py3, as `unittest2` stopped working for me while I was working on this branch.

Issue #994, opened 2015-04-07
update docs to work in Python 3


davidmarin:
All the code we supply should work in Python 2 and 3.

In the library, we can do stuff like:

```python
from mrjob.py2 import iteritems

for k, v in iteritems(some_dict):
    ...
```

but that's awkward in an example. Instead, let's just iterate through the keys of the dict, which is efficient in both Python 2 and 3:

```python
for k in some_dict:
    v = some_dict[k]
    ...
```

davidmarin:
This is pretty much done except for a couple of examples.

davidmarin:
Moving examples to Issue #1032.

Issue #989, opened 2015-04-07
Python 3 support on EMR


davidmarin:
It used to be that only `boto3` supported EMR, but now regular `boto` supports it (since v2.32.1). So it's realistic to include this in v0.4.x. Of course, we'll need to require a higher version of boto than v2.2.0 in Python 3.

We also need to automatically bootstrap install Python 3 on EMR using yum or apt-get (depending on AMI version).

davidmarin:
None of the AMIs actually have Python 3 installed on them, which makes this a bit tricky.

The 3.7.0 AMI apparently actually uses Amazon Linux 2015.03 (not 2014.03 as [the documentation](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/ami-versions-supported.html) says), so on there, we can `sudo yum install -y python34` at bootstrap time. That might be good enough for starters.

The 2.x AMIs all use Debian squeeze, which offers Python 3.1. It's probably not that hard to get mrjob working on Python 3.1, but I think this is an old enough version of Python that it would produce surprising results. Could try installing the `python34` package from a later version, but that sounds tricky.

Another option is to download the source for whatever version of Python you're running, and compile and install it. This probably has more overhead than installing a package though.

davidmarin:
The `-y` switch to `yum` is important; otherwise it just blocks forever waiting for your input.

davidmarin:
Looks like I can install `python34` without any problems on the 3.7.0 AMI, but not on the 3.3.0 AMI (which apparently uses Amazon Linux 2014.09).

davidmarin:
Downloading and installing Python from source only takes about 5 minutes, so that might be a more reliable option. Probably a better new-user experience if there's less delay, though.

davidmarin:
Here's pretty much the bootstrap script for installing from source:

```
mkdir __mrjob_python
cd __mrjob_python
wget -S -T 10 -t 5 https://www.python.org/ftp/python/x.y.z/Python-x.y.z.tgz
cd */
./configure && make && sudo make install
cd ../..
```

davidmarin:
I think our best bet is to keep things simple and fast. If you're in Python 3, the default AMI version will be 3.7.0, and we'll prepend `sudo yum install -y python34 || true` to your bootstrap commands.

If users use an earlier AMI with Python 3, we can point them to documentation about how to install any version of Python from source.

davidmarin:
[...]

PR #1007, opened 2015-05-05
Python 3.3+ compatibility

### LINKED ISSUE #830: open files in binary mode or with utf8 encoding
We should use `'rb'` or `'wb'` whenever we `open()` files, both in anticipation of supporting Python 3 (#217) and to fix some Windows compatibility issues (#296).

### LINKED ISSUE #1010: fix logging noise in tests
Python 3 actually sets up a logger, rather than just saying "no handlers for...", so it's pretty easy to figure out where tests are outputting extra logging messages. This is a good opportunity to clean up logging noise in tests.

Could bump this to a later milestone too, noisy tests are not such a big deal.

davidmarin:
This branch makes gets mrjob working in Python 3.3+ (or at least, all the tests pass), addressing #217. My general approach was to prioritize simple, readable code over wrapper functions and optimization. Please see the docstring of `mrjob/py2.py` for details.

This also restricts `unittest2` to Python 2.6 (since it's just a backport of `unittest` in Python 2.7), addressing #997. This fixes #830 as well (open files in binary mode), which will help with Windows compatibility. This also fixes #1010 (noisy logging in tests).

In the process, I found a couple of spots in the code where I thought we were iterating through a string line by line, but I was actually iterating character by character. Python 3 was the reason I caught this (when you iterate through `bytes`, on Python 3 you get `int`s, not characters).

Outside the scope of this pull request:

 * Actually supporting EMR (#989). I'm going to need to add a bootstrap action to download and compile the Python 3 source tarball. `boto` *does* support Python 3 now, so that part is done (modulo any issues that come up in hand-testing).
 * Fully updating documentation and examples to make sense on Python 3 (#994).


mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+  - use ``'b'`` when opening files: ``open(..., 'rb')``
+  - read data from ``sys.stdin.buffer`` in Python 3, not ``sys.stdin``
+- data encoded by protocols
+- data from subprocesses (this already happens by default)
+- log files parsed by mrjob
+- file content from our filesystem interfaces.
+
+Instead of using ``StringIO`` to deal with these, use ``io.BytesIO``.
+
+Not that both Python 2.6+ and Python 3.3+ have the ``bytes`` type and
```
bukzor: s/Not/Note/

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+that all non-byte strings be unicode. But that doesn't really make sense for
+Python 2, where str (bytes) and unicode can be used interchangeably.
+
+So really our string datatypes fall into two categories, bytes, and
+"strings", which are either ``str`` (i.e., bytes) or ``unicode`` in Python 2,
+and ``str`` (i.e. unicode) in Python 3.
+
+These things should always be bytes:
+
+- input data files
```
bukzor: Are data files binary blobs in general, or always some kind of text?

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+  - use ``'b'`` when opening files: ``open(..., 'rb')``
+  - read data from ``sys.stdin.buffer`` in Python 3, not ``sys.stdin``
+- data encoded by protocols
+- data from subprocesses (this already happens by default)
+- log files parsed by mrjob
```
bukzor: I'd think that log files are text, ie the canonical representation would be unicode.
Is it because you're unsure of what the encoding is?

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+
+Bytes vs. Unicode
+-----------------
+
+It's tempting to use `from __future__ import unicode_literals` and require
+that all non-byte strings be unicode. But that doesn't really make sense for
+Python 2, where str (bytes) and unicode can be used interchangeably.
+
+So really our string datatypes fall into two categories, bytes, and
+"strings", which are either ``str`` (i.e., bytes) or ``unicode`` in Python 2,
```
bukzor: In the community, we differentiate three groups, bytes (which is non-ambiguous), text (py2 unicode, py3 str), and "native" strings (str in both).

It would make more sense to me if you rephrased this doc in those terms. If I understand what's written, it looks like you're opting to use bytes and text but never native-strings.

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+This module provides a ``basestring`` type so you can check if something
+is a "string".
+
+We don't provide a ``unicode`` constant:
+
+- Use ``not isinstance(..., bytes)`` to check if a string is Unicode
+- To convert ``bytes`` to ``unicode``, use ``.decode('utf_8')`.
+- Python 3.3+ has ``u''`` literals; please use sparingly
+
+If you need to convert bytes of unknown encoding to a "string" (e.g.
```
bukzor: The "internet encoding" is for this: https://github.com/Yelp/yelp_encodings

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+
+Iterables
+---------
+
+Using ``.iteritems()`` or ``.itervalues()`` in Python 2 to iterate over a
+dictionary when you don't need a list is best practice, but it's also (in most
+cases) an over-optimization. We'd prefer clean code; just use ``.items()``
+and ``.values()``.
+
+If you *do* every need that extra efficiency ``for k in some_dict`` does not
```
bukzor: .iteritems is memory-efficient, but is not cpu-efficient.
List comprehensions and .items() always run faster in my experience.
I'd only recommend the noted strategy when you don't think you can fit another copy of the dict in memory, which I've never actually encountered.

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+create a list in either version of Python, and you for performance-critical
+code, it's fine to write custom code for each Python version (use ``IN_PY2``
+from this module to check which version you're in).
+
+Same goes for ``xrange``; just use ``range`` (this module provides ``xrange``
+solely to support ``ReprProtocol``).
```
bukzor: If you don't want xrange to be a public interface, should stick it into that one spot it's used.

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+else:
+    basestring = str
+basestring
+
+# ``long``, for ``isinstance(..., (int, long))``
+if IN_PY2:
+    long = long
+else:
+    long = int
+long
```
bukzor: You're accomplishing the same thing six does with the below code, but in a new and different way.

https://pythonhosted.org/six/#six.integer_types

I really think you should just use six instead. I understand the reluctance to add a dependency, but it's very lightweight, has no dependencies, and is *very* likely to already be installed (even yelpmain uses it).

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+
+# ``StringIO``. Useful for mocking out ``sys.stdout``, etc.
+if IN_PY2:
+    from StringIO import StringIO
+else:
+    from io import StringIO
+StringIO  # quiet, pyflakes
```
bukzor: https://bitbucket.org/gutworth/six/src/8a545f4e906f6f479a6eb8837f31d03731597687/six.py?at=default#cl-632

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+    from urllib.parse import quote
+    from urllib.parse import unquote
+    from urllib.request import urlopen
+    from urllib.parse import urlparse
+    from urllib.parse import ParseResult
+quote
+unquote
+urlopen
+urlparse
+ParseResult
```
bukzor: https://bitbucket.org/gutworth/six/src/8a545f4e906f6f479a6eb8837f31d03731597687/six.py?at=default#cl-320

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+def to_string(s):
+    """On Python 3, convert ``bytes`` to ``str`` (i.e. unicode).
+
+    (In Python 2, either ``str`` or ``unicode`` is acceptable as a string.)
+
+    Use this if you need to ``print()`` or log bytes of an unknown encoding,
+    or to parse strings out of bytes of unknown encoding (e.g. a log file).
+    """
+    if not isinstance(s, (bytes, str)):
+        raise TypeError
```
bukzor: You said str is acceptable, above.
This error could really use a message.

mrjob/py2.py
```diff
@@ -0,0 +1,170 @@
...
+
+    if IN_PY2 or isinstance(s, str):
+        return s
+
+    try:
+        return s.decode('utf_8')
+    except UnicodeDecodeError:
+        return s.decode('latin_1')
```
bukzor: yelp's from_bytes is this function: https://github.com/Yelp/yelp_bytes

mrjob/conf.py
```diff
@@ -68,8 +67,8 @@ def default_options(self):
     def validated_options(self, opts, error_fmt):
         unrecognized_opts = set(opts) - self.ALLOWED_KEYS
         if unrecognized_opts:
-            log.warn(error_fmt % ', '.join(sorted(unrecognized_opts)))
-            return dict((k, v) for k, v in opts.iteritems()
+            log.warning(error_fmt % ', '.join(sorted(unrecognized_opts)))
```
bukzor: pylint would give you a warning here: non-lazy log
The log functions take two arguments, a %-style template, and values for the template.
It defers actually templating the stuff until it decides to emit the logline.

P4

mrjob/emr.py
```diff
@@ -2454,7 +2450,14 @@ def _pool_hash(self):
 
         if self._opts['bootstrap_mrjob']:
             things_to_hash.append(mrjob.__version__)
-        return hash_object(things_to_hash)
+
+        things_json = json.dumps(things_to_hash, sort_keys=True)
+        if not isinstance(things_json, bytes):
+            things_json = things_json.encode('utf_8')
```
bukzor: Uh, the canonical name of that encoding is "UTF-8".
The uppercase actually saves some nanoseconds..

Also, if you do dep yelp-bytes, this is the to-utf8 function.

mrjob/hadoop.py
```diff
@@ -333,12 +334,12 @@ def _run_job_in_hadoop(self):
...
-                    # stderr and stdout (it's a fake terminal)
-                    self._process_stderr_from_streaming(master)
-                    _, returncode = os.waitpid(pid, 0)
-                    master.close()
+                    with os.fdopen(master_fd, 'rb') as master:
+                        # reading from master gives us the subprocess's
+                        # stderr and stdout (it's a fake terminal)
+                        self._process_stderr_from_streaming(master)
+                        _, returncode = os.waitpid(pid, 0)
+                        master.close()
```
bukzor: The context-manager closes the file.
This is a double-close.

p3

mrjob/py2.py
davidmarin: They're always binary blobs.

mrjob/py2.py
davidmarin: Yeah. Considering log files to be "hostile" (i.e. their contents out of our control), and want mrjob to deal gracefully with weird encodings. But yeah, they're going to be produced by Java (Hadoop), so in principle, they're unicode.

I mean, probably 99.9% of the time they're going to be purely ASCII, so encoding isn't even going to matter.

bukzor: internet-encoding is for this: https://github.com/Yelp/yelp_encodings

It will bring your three-nines to six-nines and never error.
In that 1/1000000 case, you'll still get unicode, just encoded slightly wrong.

mrjob/py2.py
davidmarin: That makes sense for a new library or internal code.

The problem is that Python 2 is pretty casual about bytes vs. unicode, and it needs to continue to work that way, or Python 2 users will run into unexpected issues. That's why I divided things up this way.

mrjob/py2.py
bukzor: To be clear, I'm only suggesting you change the terminology, not anything else. "text" is unambiguous where your usage of "strings" is unusual.

mrjob/py2.py
davidmarin: Neat, makes sense that Yelp would want to try Windows encoding before falling back to latin-1.

Don't anticipate seeing a lot of smart quotes in log files though. Again, most of this stuff is probably plain ASCII anyway.

mrjob/py2.py
bukzor: To be clear, I'm only suggesting you change the terminology, not anything else. "text" is unambiguous where your usage of "strings" is unusual.

mrjob/job.py
```diff
@@ -67,6 +63,20 @@
...
+
+    Python 2 is slightly different because classes have "unbound methods"
+    which wrap the underlying function, whereas on Python 3 they're just
+    functions. (Methods work the same way on both versions.)
+    """
+    # "im_func" is the old Python 2 name for __func__
+    if hasattr(f, '__func__'):
+        return f.__func__
+    else:
+        return f
```
bukzor: https://bitbucket.org/gutworth/six/src/8a545f4e906f6f479a6eb8837f31d03731597687/six.py?at=default#cl-536

mrjob/py2.py
davidmarin: Huh! Should have done some timing tests, I guess.

Fortunately, this never came up; the only performance-critical parts of mrjob are protocols and the `MRJob` methods responsible for running the job, and they don't want to use `iteritems()` anyway.

mrjob/py2.py
davidmarin: Good point, will do.

mrjob/py2.py
davidmarin: mrjob has *no* required dependencies, want to keep it that way.

mrjob/py2.py
davidmarin: Neat. Windows encoding is pointless for mrjob, but `from_bytes()` is a better name, will use that.

mrjob/py2.py
bukzor: In that case I'd urge you to replace mrjob.py2 with mrjob.six.
Six is a single file with no dependencies.

mrjob/conf.py
davidmarin: Looks like I'd have to change >100 lines to clean this up, so not in this pull request. Want me to file a "cleanup" issue on it?

mrjob/emr.py
davidmarin: Huh, really? Was trying to match [Standard Encodings](https://docs.python.org/3/library/codecs.html#standard-encodings) from the Python docs.

Yeah, was considering having a function for this, but it only happens like six times in the code base, so I decided to err on the side of readability (less indirection). Whereas `to_string()` actually *does* something, so I relented and made it a function.

mrjob/hadoop.py
davidmarin: Thanks! Will fix.

mrjob/py2.py
bukzor: It's not pointless if any of these bytes are related to the internet.
cp1252 is what all browsers use in place of latin1, and the WHATWG has formalized it in their encoding spec: https://encoding.spec.whatwg.org/#names-and-labels
It's why I named it internet-encoding.

mrjob/conf.py
bukzor: Very low priority (p4). Whatever you like.

mrjob/job.py
davidmarin: Again, not going to add `six` as a dependency, but this is one place where I think mine is a little bit better; it's simple, and it can handle methods, unbound methods, and regular functions.

This function is really only used in one place (two, if you count tests), which is why I left it in `job.py`.

mrjob/py2.py
davidmarin: How should I change the comment to make it more clear?

mrjob/py2.py
davidmarin: Wait, will my code error? I thought falling back to latin-1 is pretty much bulletproof.

mrjob/py2.py
davidmarin: Got it. You're right, "text" is better.

mrjob/py2.py
davidmarin: Eh, it's still a ton of unnecessary code. `six` was designed with, like, bridging the gap between Python 2.4 and Python 3.0. My feeling is that a lot of it is obsolete for newer versions of Python.

Normally, I'd buy the argument that my code is new and untested, whereas `six` has been around a while, but my code doesn't *do* anything (other than `to_string()`).

hayd:
Can you add python 3.3 and 3.4 as supported in setup.py?

Both [2.6](https://travis-ci.org/davidmarin/mrjob/jobs/61361961) and [3.4](https://travis-ci.org/davidmarin/mrjob/jobs/61361972) splutter some warnings (although they pass the tests).

mrjob/py2.py
davidmarin: Will change the comment accordingly.

mrjob/py2.py
davidmarin: [ ] Will do.

mrjob/py2.py
bukzor: It's mainly that I'd like newcomers to find familiar terminology and methods.
You can delete the irrelevant parts if you like.
What you'd be left with is your code in community-standard form.

mrjob/emr.py
bukzor: Looks like you were right to start with:

```python
>>> encodings.search_function('UTF-8')
>>> encodings.search_function('utf_8')
<codecs.CodecInfo object for encoding utf-8 at 0x1088ba940>
>>> encodings.search_function('UTF_8')
>>>
```

mrjob/py2.py
davidmarin: Oh, I see. Yeah, I should at least use the same names as `six`

 - [ ] use `integer_types` instead of `(int, long)`
 - [ ] use `string_types` instead of `basestring`

mrjob/py2.py
bukzor: Latin-1 is bullet-proof in the sense that it will never except, but it will likely be wrong if a non-ascii character appears. internet-encoding preserves the no-except behavior but tries utf-8 first.

This stuff doesn't matter much, just fyi.
Do what you want to maintain :)

mrjob/py2.py
davidmarin: Very well then.

 - [ ] rename `to_string()` to `from_bytes()`
 - [ ] handle cp1252 encoding in log files/subprocesses
 - [ ] add automated test for this

mrjob/py2.py
davidmarin: Oh, and:

 - [ ] rename `IN_PY2` to `PY2`

mrjob/py2.py
davidmarin: (Not sure how this is different from the above comment.)

mrjob/py2.py
bukzor: cp1252 as python implements it is flawed though: it will except if given any of five non-printable bytes. The whatwg specifies otherwise but I wasn't able to convince unicode.org to change their spec =/
It means a third and final fallback of latin1 is still necessary.
This is exactly what unicode-encoding implements.

mrjob/py2.py
davidmarin: `six`'s approach is pretty heavyweight (don't want a `mrjob.py2.moves` module). Just referring to these functions without their module name works fine for mrjob.

davidmarin: Err, that is, I think it's a good idea to match the names `six` uses rather than making up my own. But it's even better to just match the names in the Python standard library.

mrjob/emr.py
davidmarin: **woot**

davidmarin:
Was going to update `setup.py` once Python 3 really works with EMR, but, sure, why not:

 [ ] - Add Python 3.3/3.4 to `setup.py`

I have a ticket for the `ResourceWarning`s (#1009) in Python 3, and to quiet logging messages (#1010). Want me to file a ticket for the `atexit` warnings in Python 2.6? Anything else I missed?

mrjob/py2.py
davidmarin: Kind of surprised that's the only typo!

mrjob/conf.py
davidmarin: Okay, leaning towards fail-fast on bad log formatting strings (i.e., the way we do it now). We don't log enough stuff that there's any question of efficiency.

mrjob/py2.py
davidmarin: Oh, I see, you're suggesting opening log files as unicode, in `internet` encoding. That *does* make sense; I agree that log files are meant to be unicode, not binary blobs.

Here's the issue: mrjob actually gets most of its log data through `cat()` in its "filesystem" interface, which is just a generator of `bytes` (not a filehandle). So it would be a fair amount of code in addition to Yelp's codec code to make that work. I can't just wrap it in `codecs.getreader()`.

So I'm going to just use `from_bytes()` instead. I *will* add cp1252 encoding (see below).

Also, will probably want to use `yelp_encoding` for my startup's code; don't mind extra deps there.

Did you notice that the docstring for [internet.py](https://github.com/Yelp/yelp_encodings/blob/master/yelp_encodings/internet.py) is incorrect? It doesn't mention cp1252, and I'm not sure what the remark about "latin1html5" means.

mrjob/runner.py
```diff
@@ -227,15 +223,27 @@ def validate_cleanup(error_str, opt_list):
...
+
+        This returns a single-item list (because it's a command).
+        """
+        if local and sys.executable:
+            return [sys.executable]
+        elif IN_PY2:
+            return ['python']
+        else:
+            # e.g. python3
+            return ['python%d' % sys.version_info[0]]
```
bukzor: In what case do you not want sys.executable?

davidmarin: When the code is going to execute on another host (i.e. in Hadoop/EMR).

mrjob/py2.py
davidmarin: Oh, okay, if you're willing to let cp1252 slide, so am I. Just doesn't seem that relevant to Hadoop.

mrjob/py2.py
davidmarin: Actually, "from_bytes()" is a little confusing because it also allows the input to be unicode. I think I should rename it to `to_text()` to match using the term "text" instead of "string".

davidmarin: - [ ] rename `to_string()` to `to_text()`
 - [ ] write automated tests for `to_text()`

mrjob/job.py
```diff
@@ -400,25 +406,26 @@ def increment_counter(self, group, counter, amount=1):
         if not isinstance(amount, (int, long)):
             raise TypeError('amount must be an integer, not %r' % (amount,))
 
+        # cast non-strings to strings (if people pass in exceptions, etc)
+        if not isinstance(group, basestring):
+            group = str(group)
+        if not isinstance(counter, basestring):
+            counter = str(counter)
+
```
bukzor: This stuff would be less weird (and have one less if-isinstance) if you invariantly made sure these things were str.

mrjob/util.py
```diff
@@ -611,7 +621,13 @@ def safeeval(expr, globals=None, locals=None):
...
     # blank out builtins, but keep None, True, and False
     safe_globals = {'__builtins__': None, 'True': True, 'False': False,
-                    'None': None, 'set': set, 'xrange': xrange}
+                    'None': None, 'set': set}
+
+    # xrange is range in Python 3
+    if hasattr(__builtins__, 'xrange'):
+        safe_globals['xrange'] = xrange
+    else:
+        safe_globals['range'] = range
```
bukzor: should keep range in both cases?

tests/sandbox.py
```diff
@@ -118,8 +130,11 @@ def add_mrjob_to_pythonpath(self):
...
-
         os.environ['PYTHONPATH'] = (
-            mrjob_pythonpath + ':' + os.environ.get('PYTHONPATH', ''))
+            mrjob_pythonpath() + ':' + os.environ.get('PYTHONPATH', ''))
+
+
+def mrjob_pythonpath():
+    """The directory containing the mrjob package that we've imported."""
+    return os.path.abspath(
+        os.path.join(os.path.dirname(mrjob.__file__), '..'))
```
bukzor: I think dirname and join-".." are the same thing.

p4

tests/suite.py
```diff
@@ -1,12 +1,13 @@
...
+from os.path import dirname
 
-try:
+# this module has to stand alone, so we can't use tests.py2
+if sys.version_info < (2, 7):
     import unittest2 as unittest
-    unittest  # quiet "redefinition of unused ..." warning from pyflakes
-except ImportError:
+else:
     import unittest
```
bukzor: reuse what you've put in tests.py2?

tests/test_protocol.py
```diff
@@ -41,11 +36,14 @@ def __repr__(self):
...
+    def __eq__(self, other):
         if not isinstance(other, Point):
-            return 1
-        else:
-            return cmp((self.x, self.y), (other.x, other.y))
+            return False
+
+        return (self.x, self.y) == (other.x, other.y)
+
+    def __ne__(self, other):
```
bukzor: You don't need total ordering?

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py26, py27, pypy
+envlist = py26, py27, pypy, py33, py34
```
bukzor: There's also pypy3, but maybe that's for another branch.

mrjob/py2.py
bukzor: That last sentence is wayy too complex.

It might be best to discuss in person.

I take issue with "Python 2, where str (bytes) and unicode can be used interchangeably". This is *only* true for English strings.

mrjob/emr.py
bukzor: Profiling shows that 'utf-8' can be twice as fast as 'UTF-8' or 'utf_8' in json-heavy applications.

mrjob/py2.py
bukzor: You're correct, that docstring has rotted.
Filed https://github.com/Yelp/yelp_encodings/issues/4

latin1html5 was meant to refer to the whatwg's specification of latin1: cp1252 but with the five "unknown" bytes decoding to unprintable characters of equal ordinal.

tests/sandbox.py
davidmarin: Nope, you can't put `..` after a filename, has to be a directory name.

mrjob/job.py
davidmarin: This is a carryover from the old Python 2 code, which allows you to use unicode for counter/group names.

mrjob/util.py
davidmarin: Nah, the point of `safeeval()` is to decode reprs, which won't have `range` in them on Python 2, since in Python 2, `range()` is just a function that returns a list of numbers.

By the way turns out this code was wrong; `__builtins__` is actually a dictionary on Python 2 (it supports attribute access, but not `hasattr()`). Good thing I added a test case!

tests/sandbox.py
bukzor: That doesn't make sense, to me. These things are strictly string munging, with no filesystem access. In this example, bar  is neither a file nor a directory:

```python
>>> os.path.abspath(os.path.join('/foo/bar', '..'))
'/foo'
```

Anyhow, not important.

tests/suite.py
davidmarin: It doesn't seem to be able to find `tests.py2` when run from `setup.py`. Not sure how that works exactly, but didn't want to wrestle with it.

bukzor: "when run from setup.py"? If you mean the setuptools test command, that thing is hopelessly broken.

tests/test_protocol.py
davidmarin: Nope, just equality. Python 3 is much more bare-bones; there's no `__cmp__()` (you have to use `__gt__()`, `__lt__()`, etc.), and it won't automatically infer `__ne__()` from `__eq__()`.

tox.ini
davidmarin: Oh, so there is! It's Python 3.2 compatible though.

davidmarin:
Sounds good.

tox.ini
davidmarin: Ah, but with `u''` literals. That's probably good enough.

 - [ ] add pypy3 test target

tests/test_util.py
```diff
@@ -280,8 +280,19 @@ def test_globals_and_locals(self):
             abs(a),
             safeeval('abs(a)', globals={'abs': abs}, locals={'a': a}))
 
+    def test_range_type(self):
+        if PY2:
+            # on Python 2, the range type is called xrange, and doesn't
+            # support equality
+            self.assertEqual(repr(safeeval('xrange(3)')), 'xrange(3)')
+        else:
+            self.assertEqual(safeeval('range(3)'), range(3))
```
bukzor: if you change this as below, pypy3 will pass all tests:

```python
self.assertEqual(safeeval('tuple(range(3))'), (0, 1, 2))
```

davidmarin:
Huh, well, I *thought* that would fix it, but there is another issue on `pypy3`. And there's too much logging noise in the tests (#1010) for me to figure out what's wrong.

My inclination would be to merge this in now, and make support for `pypy3` a separate ticket, which I'll plan to fix ASAP along with #1010. If that's okay with you, please give me a "Ship It". Otherwise, I'll try to attack this when I have time tomorrow.

@buzkor, how *did* you figure out that the range test was responsible for the failure on pypy3? Is there a secret Travis CI feature that helps you sort through the noise?

davidmarin:
Actually, never mind, fixing #1010 was easy. Let's see if this works.

bukzor:
I was looking at run 552.6, where the summary is pretty clear.

:shipit:

bukzor:
Since you don't in fact test range() repr's anymore, should you remove it's support?

davidmarin:
Cool.

There's an explicit test for the range type later in that test case. Removing support would break backwards compatibility.

PR #1020, opened 2015-05-24
Python 3 on EMR

### LINKED ISSUE #989: Python 3 support on EMR

### LINKED ISSUE #992: default instance type to m1.medium
We should pick an appropriate instance type when running newer AMIs.

Apparently, asking EMR for the "latest" version gives us AMI version 2.4.2, even though versions 2.4.11 and 3.6.0 exist (see [AMI list](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/ami-versions-supported.html)). Good thing, or mrjob would be broken out of the box!

### LINKED ISSUE #1014: default to AMI 3.11.0
EMR is phasing out "latest" (which is stuck somewhere in the 2.x AMI series). `ami_version` should default to 3.7.0, which is currently the actual latest AMI and the only one where you can install Python 3 from a package.

davidmarin:
These are the final changes needed to make mrjob work on EMR in Python 3 (fixes #989). They are:

 * change the default AMI to 3.7.0 (fixes #1014)
 * change the default instance type to `m1.medium` (fixes #992)
 * In Python 3, run `sudo yum install -y python34` in the master bootstrap script

There's a `bootstrap_python` option that you can set to false if you want to install Python 3 some other way; I updated the docs to reflect how to do this from source.

Hand-tested on Python 2 and 3.

`master` now targets v0.5.0, so it's okay to change defaults.


bchess:
Lg2m, ship it

Issue #1008, opened 2015-05-05
fall back to default profile/role if we can't access IAM


davidmarin:
mrjob can auto-create the roles/profiles that EMR requires, but this only works if you can access IAM.

For users with more limited access (such as the OP of [this Google Group thread](https://groups.google.com/forum/#!topic/mrjob/Aurb85zV2s4)), probably the best we can do is assume that somebody ran `aws emr create-default-roles` (as suggested [here](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-iam-roles-defaultroles.html)), and if not, to suggest they point their administrator to that page.

PR #1029, opened 2015-05-26
Fallback IAM role/profile

davidmarin:
This implements #1008.

Basically, if someone uses mrjob for the first time, and they have full IAM access, mrjob can just create the service role and instance profile for them.

However, if they're don't have full access to IAM, this approach won't work. In this case, the best we can do is hope they (or someone on their account) has followed [Amazon's instructions on how to create these](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-iam-roles-creatingroles.html), and just go ahead and try to use them. If this fails (because they haven't been created after all), we give them a URL to show to their admin.

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -1394,8 +1394,9 @@ class EMRJobRunner(MRJobRunner):
         except boto.exception.BotoServerError as ex:
             if ex.code != 'AccessDenied':
                 raise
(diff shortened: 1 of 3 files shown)


davidmarin:
Whoops, meant to submit this against the `v0.4.x` branch.

PR #1030, opened 2015-05-26
Fallback IAM role/profile

davidmarin:
This implements #1008.

Basically, if someone uses mrjob for the first time, and they have full IAM access, mrjob can just create the service role and instance profile for them.

However, if they're don't have full access to IAM, this approach won't work. In this case, the best we can do is hope they (or someone on their account) has followed [Amazon's instructions on how to create these](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-iam-roles-creatingroles.html), and just go ahead and try to use them. If this fails (because they haven't been created after all), we give them a URL to show to their admin.

This is against the `v0.4.x` branch because it's going to go into `v0.4.5`, which will be a handful of IAM fixes plus some deprecation warnings. As soon as I merge this request, I'll merge `v0.4.x` into `master` (which currently targets `v0.5.0`).

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -1394,8 +1394,9 @@ class EMRJobRunner(MRJobRunner):
         except boto.exception.BotoServerError as ex:
             if ex.code != 'AccessDenied':
                 raise
-            log.warning('Trying fallback instance profile: %s' %
-                        FALLBACK_INSTANCE_PROFILE)
+            log.warning(
+                "Can't access IAM API, trying default instance profile: %s" %
+                FALLBACK_INSTANCE_PROFILE)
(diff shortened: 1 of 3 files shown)


anusha-r:
Looks good to me! Do you want me to merge it in?

davidmarin:
Feel free to merge it!

Enviado con Aquamail para Android
http://www.aqua-mail.com


El 26 de mayo de 2015 9:36:45 AM Anusha Rajan <notifications@github.com> 
escribio:

davidmarin:
Or whatever. But yeah, if you're the reviewer, and you have commit privileges, go for it! Nothing says "looks good to me" like actually merging the code. :)

Issue #1026, opened 2015-05-24
use built-in managed IAM policies


davidmarin:
Currently, mrjob hard-codes the policies it creates when auto-creating default IAM roles/policies/instance profiles so that mrjob can work out-of-the-box.

The policies, at least, already exist on EMR as `AmazonElasticMapReduceRole` and `AmazonElasticMapReduceforEC2Role`. We should just attach these policies to the roles we create, rather than creating our own.

This simplifies the code, but it also allows us to automatically keep up with changes to these policies.

davidmarin:
The roles and instance profile are pretty much just simple wrappers, and unlikely to change.

davidmarin:
Doing this cleanly is actually a little more complicated than what we have right now. For example, for service role, what we *really* want is a role that uses `AmazonElasticMapReduceRole` as its policy. But a role that has an identical policy (e.g. the one created by the AWS CLI) would also be fine.

PR #1031, opened 2015-05-30
use managed IAM policies

davidmarin:
This implements #1026.

Basically, we're working a little harder than we need to to set up IAM for EMR. We *do* need to create roles and an instance profile, but instead of creating policies, we can simply "attach" ones already in AWS. This is not only simpler, it's better because these "managed" policies are versioned and will be automatically updated to track changes in EMR.

Apparently (based on email conversations with a senior manager on the EMR team), the AWS web interface and CLI are going to work this way soon as well. Once they do, mrjob will be able to find and use the role/profile they create rather than auto-creating its own.

Hand-tested this in addition to updating the unit tests.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,7 +1,9 @@
-v0.4.5, 2015-??-?? -- ???
+v0.4.5, 2015-??-?? -- IAM Fortified
  * runners:
    * EMR:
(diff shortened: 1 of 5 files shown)


bchess:
This looks good to me

Issue #1036, opened 2015-06-04
RawValueProtocol incompatible with JSONProtocol in Python 3


davidmarin:
By default, jobs read in input with `RawValueProtocol` which returns `(None, line_as_bytes)`.

On Python 2, `json.dumps()` will accept bytes (as long as they're UTF-8 encoded), so something like this will work fine:

```python
class MRWordFreqCount(MRJob):

    def mapper(self, _, line):
        for word in WORD_RE.findall(line):
            yield (word.lower(), 1)

    def reducer(self, word, counts):
        yield (word, sum(counts))
```

In Python 3, `line` is bytes, which means the words are bytes, which means `json.dumps()` won't accept them.

What we should do is have a `BytesValueProtocol` to read bytes, and a `TextValueProtocol` to convert bytes to unicode (first trying UTF-8, and falling back to latin-1). On Python 2, `RawValueProtocol` will be an alias for `BytesValueProtocol`, and on Python 3, `RawValueProtocol` will be an alias for `TextValueProtocol`.

We should do the same for `RawProtocol` (and `BytesProtocol`, `TextProtocol`) for consistency.

davidmarin:
`to_string()` might be a bit lenient; UTF-8 decoding might be closer to what we have now. In Python 2, if a job reads in non-UTF-8 bytes, it won't be able to JSON-encode them anyway.

davidmarin:
Another approach would be to have `BytesValueProtocol` to just read bytes, and `TextValueProtocol` to convert bytes to unicode (first trying UTF-8, and falling back to latin-1). Then we can make `RawValueProtocol` an alias for `BytesValueProtocol` on Python 2, and `TextValueProtocol` on Python 3.

Issue #1002, opened 2015-04-15
ujson in protocols, json elsewhere


davidmarin:
Now that mrjob is a 2.6+ library, the `json` module always exists. For places where performance doesn't matter (e.g. reading config files), we should just use `json` for simplicity.

It looks like `simplejson` is no longer that fast (sometimes slower than `json`, see #1006). For places where performance *does* matter (that is, protocols), we should be using [ujson](https://pypi.python.org/pypi/ujson).

davidmarin:
Note that `ujson` doesn't do circular object checks at all; it'll just give an `OverflowError`:

```
>>> a = {}
>>> a['a'] = a
>>> ujson.dumps(a)
OverflowError: Maximum recursion level reached
```

That's fine, especially now that we're moving to strict protocols by default (see #724).

davidmarin:
`ujson` can also accept `bytes` in Python 3, which `json` can't. Should leverage this, since performance matters here.

davidmarin:
Like we're planning to do with Text/Bytes protocols in #1036, there should ba a `StandardJSONValueProtocol` and an `UltraJSONValueProtocol`, with `JSONValueProtocol` aliasing to one or the other.

Issue #724, opened 2013-09-03
make protocols strict by default


davidmarin:
I'm guessing hardly anyone paid attention to [our gentle warning](http://mrjob.readthedocs.org/en/latest/whats-new.html#id2) to stop using loose protocols. While it would be nice to eliminate them entirely, it's more practical to simply make them no longer the default.

When someone does turn loose protocols on, we can warn that they'll be removed entirely in v0.6.0.

PR #1040, opened 2015-06-12
Protocol revamp

davidmarin:
Three interrelated changed to protocols:

**JSON** (issue #1002):

This rips out `simplejson` (which is apparently not much faster or even slower than the built-in implementation) and replaces it with `ujson` (which is a lot faster).

The way things used to work is that `JSONProtocol` (and the rest of mrjob) would use `simplejson` if it is available, and regular `json` otherwise.

`ujson` is different enough from built-in `json` that we make separate `Standard` and `Ultra` JSON protocols available, and make `JSONProtocol` an alias for the fastest one available. We do the same for `JSONValueProtocol`.

We also simplified the rest of the code to just use the built-in `json` module; there's no performance advantage, to, say, using `ujson` to read `mrjob.conf`.

**Strict protocols** (issue #724):

This has been a long time in coming. Currently if a protocol can't decode a line or encode a key-value pair, mrjob increments a counter instead of raising an exception. This is great for, say, log files with a sprinkling of non-UTF-8 bytes, but confusing and awful if you use the wrong protocol or have a bug in your protocol implementation, and ignores most or all of your data.

This change turns off this "feature". You can still enable it with the `--no-strict-protocols` switch or setting `strict_protocols: false` in `mrjob.conf`.

We're also planning to eventually make protocols able to handle non-line-based data. In this case, the protocols, rather than the `MRJob` class will be responsible for determining where one record ends and another begins, and so "loose" protocols won't really make sense.

**Read input as `str`** (issue #1036):

It took me a while after the Python 3 port to realize that `mrjob.example.mr_word_freq_count` didn't work in Python 3!

What happened is that the job would use `RawValueProtocol` (the default) to read in lines as `bytes`, but would use a `str` regex to parse them. Even if it were using a `bytes` regex, however, Python 3 wouldn't JSON-encode them! (`mr_word_freq_count` now works in Python 3, and has a regression test.)

`RawValueProtocol` is now an alias for `BytesValueProtocol` on Python 2 and `TextValueProtocol` on Python 3. `TextValueProtocol` reads in lines as unicode strings (`str` on Python 3, `unicode` on Python 2); it hopes that lines will be UTF-8 decodable, but falls back to latin-1.

Because of this fallback `TextValueProtocol` is actually a little better than `BytesValueProtocol`. If your mapper reads a non-UTF-8-encoded line from a log file with `TextValueProtocol` and then yields it, `JSONProtocol` will always be able to handle it, whereas `JSONProtocol` won't be able to handle non-UTF-8 bytes read by `BytesValueProtocol`.

I think this issue may be why I created "loose" protocols were created in the first place. I'm not making `TextValueProtocol` the default for Python 2 because it would massively break backwards compatibility, but it's probably a good idea to set `INPUT_PROTOCOL = TextValueProtocol` in jobs that read messy log files.


.gitignore
```diff
@@ -1,5 +1,5 @@
 .#*
-#*#
+\#*#
```
bchess: Whats this about?

setup.py
```diff
@@ -21,15 +22,18 @@
     # arguments that distutils doesn't understand
     setuptools_kwargs = {
         'install_requires': [
-            'filechunkio',
             'PyYAML',
-            'simplejson>=2.0.9',
+            'filechunkio',
```
bchess: nit: maintain alphabetic ordering, keep filechunkio above PyYaml

setup.py
```diff
@@ -21,15 +22,18 @@
...
         ],
         'provides': ['mrjob'],
         'test_suite': 'tests.suite.load_tests',
         'zip_safe': False,  # so that we can bootstrap mrjob
     }
 
+    # ujson is recommended, but requires CPython
+    if platform.python_implementation() == 'CPython':
+        setuptools_kwargs['install_requires'].append('ujson')
```
bchess: Can this be specified as an "extras_require" instead?

bchess:
Looks good in general. Mostly minor comments above

.gitignore
davidmarin: Just a plain `#` starts a comment, so this wasn't catching files like `#emr.py#` created by emacs.

setup.py
davidmarin: I actually did run `M-x sort-lines` on these. :) Remember, capital letters have lower codepoints than lowercase ones.

setup.py
davidmarin: Will do. I've tried to avoid this in general (even `boto` is "optional" in the sense that mrjob can run without it if you don't use EMR), but `ujson` requires a compiler, which might be a showstopper for a lot of people.

davidmarin:
Okay, let's make sure tests still pass. Assuming they do, shall I merge it in?

bchess:
Sounds good to me!

Issue #1041, opened 2015-06-18
don't bootstrap mrjob if interpreter is set


davidmarin:
The `interpreter` option is for non-Python binaries. If we're using an `interpreter`, there's no real reason to bootstrap mrjob.

Issue #1038, opened 2015-06-05
steps_python_bin should not default to python_bin


asottile:
This says it falls back to sys.executable:
http://mrjob.readthedocs.org/en/latest/guides/configs-all-runners.html#option-steps_python_bin

The code says otherwise:
https://github.com/Yelp/mrjob/blob/4a66b1c0ee12d540b2e3d6f43391cae6b9bc6ca1/mrjob/runner.py#L241-L243

davidmarin:
Yep, good catch. The docs are out-of-date.

davidmarin:
Oh, no, never mind. Look carefully at the code. If `local=True` (which it does for `steps_python_bin`), `_default_python_bin()` does indeed return `[sys.executable]`.

asottile:
@davidmarin correct.  But the code I highlighted says `self['python_bin'] or self._default_python_bin(local=True))`

which is making it quite difficult when I want my remote python to be a virtualenv's python and I want my local python to be sys.executable

davidmarin:
Ah, I see, you want to be able to set `python_bin` without changing `steps_python_bin`. That sounds reasonable. Because it changes the defaults, I'm going to save this for v0.5.0 (which will come out about the same time as v0.4.5 anyway).

asottile:
It only changes the defaults if you're modifying python_bin so it probably has less of an impact, but sure 5.0 will be fine.

For now I'm working around it with this:

```yaml
    python_bin: /home/hadoop/virtualenv_run/bin/python
    # Hack until https://github.com/Yelp/mrjob/issues/1038
    steps_python_bin: python
```

davidmarin:
Yep, that is the workaround. Honestly, v0.4.5 mostly exists to get in those last few deprecation warnings, so it'll be coming out about the same time as v0.5.0.

And yeah, debatably this is a bug because it contradicts the docs. I just remember having what I thought was a good reason for making it work that way when when I wrote that code. It's really more of a suboptimal design decision, and that should wait for a major release.

davidmarin:
Aha! This goes back to when we were introducing partial support for scripts in other languages to mrjob. If you wanted to run, say, a job written in Ruby, you'd set `interpreter` to `ruby`, and `steps_interpreter` would also be set to `ruby` automatically.

This makes less sense for `python_bin` and `steps_python_bin` (because `sys.executable` is almost always right), but that's the reasoning.

davidmarin:
Working on this now.

PR #1043, opened 2015-06-18
python_bin and interpreter options

davidmarin:
This change decouples `steps_python_bin` from `python_bin`, as requested in #1038.

This also explicitly un-deprecates the `python_bin` and `steps_python_bin` option, which we thought at one time would be superseded by the `interpreter` and `steps_interpreter` options, and makes explicit that these options are only meant for non-Python binaries (bet you didn't know you could write a job in bash!).

Part of the reason to decouple `python_bin` and `interpreter` is that they have different defaults. If we're not using Python, we probably want to use `interpreter` both to read steps from the job locally, and to run it remotely. Whereas, if we *are* using Python, you can't really beat `sys.executable` for running the job locally.

Since setting `interpreter` means you're not using Python, doing so now implicitly turns off bootstrapping of mrjob unless you set `bootstrap_mrjob` explicitly (see #1041).

### DIFF
--- a/docs/guides/configs-all-runners.rst
+++ b/docs/guides/configs-all-runners.rst
@@ -201,7 +201,8 @@ Job execution context
     :default: ``None``
 
     Interpreter to launch your script with (e.g. ``'ruby'``). Change this if
-    you're using a language other than Python.
+    you're using a language other than Python. This will also be used to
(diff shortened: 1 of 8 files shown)


asottile:
:shipit:

davidmarin:
Thanks!

Issue #980, opened 2015-03-17
require boto 2.35.0; remove S3, EMR workarounds


davidmarin:
The current version of boto is 2.38.0, but they haven't actually changed the EMR, S3, or IAM code since 2.35.0.

This would allow us to rip out a lot of workarounds for old versions of boto:
 * `mrjob/emr.py`:
   * can always set `security_token` for both EMR and IAM
   * HMAC v4 authentication workaround
 * `mrjob/fs/s3.py`:
   * `_get_bucket()` wrapper

davidmarin:
Fixed by #1045.

davidmarin:
Whoops, set this as the default, but still need to clean up the hacks for old versions of boto.

davidmarin:
broke out IAM workarounds into a separate ticket (#1062)

PR #1045, opened 2015-06-18
set visible_to_all_users to True by default

davidmarin:
The long-awaited change to visible_to_all_users (see #930).

This also bumps the version for boto (see #980), which is part of the plan for v0.5.0. I was hoping to use `run_jobflow()`'s new `visible_to_all_users` keyword, but this would keep the `emr_api_params` option from overriding `VisibleToAllUsers` as suggested by the mrjob documentation (boto sort of broke my `api_params` patch).

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -236,14 +236,22 @@ Job flow creation and configuration
 
 .. mrjob-opt::
     :config: visible_to_all_users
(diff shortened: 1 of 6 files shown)


anusha-r:
Looks good! Will wait for the build to complete before merging in :)

davidmarin:
Thanks! I got it.

PR #1048, opened 2015-06-22
Use Instances.Ec2SubnetId in doc examples

tylernisonoff:
The correct parameter to set the EMR job in a specific subnet is Instances.Ec2SubnetId, not Instance.Ec2SubnetId.  While the docs do not explicitly say this is the intention of the parameter, it can be misleading to readers as the correct parameter to pass for this behavior. Instance.Ec2SubnetId gets ignored by the Ec2 API and has no effect.  

I've also updated the tests to reflect the change, since it seemed like the tests and docs were mirroring each other.

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -140,7 +140,7 @@ Job flow creation and configuration
 
     .. code-block:: sh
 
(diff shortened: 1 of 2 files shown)


davidmarin:
Thanks!

PR #1051, opened 2015-06-24
Syntax Highlighting in Readme.rst

tianhuil:


davidmarin:
Wish I could, but as well as displaying on GitHub, the README has to be compiled by whatever old version of `rst2html` PyPI is running.

Issue #819, opened 2013-11-15
buffer_iterator_to_line_iterator() shouldn't add final newline


davidmarin:
If a file doesn't end with a newline, the last "line" yielded by a file object won't end with a newline either.

`mrjob.util.buffer_iterator_to_line_iterator()` should act the same way.

This is a one-line change, plus possibly some test changes.

davidmarin:
Also should rename this to `yield_lines()` so that it's not so awkward to use.

Issue #715, opened 2013-08-29
rework protocols to handle non-line-based data


davidmarin:
We've gotten a lot of mail on the [mrjob mailing list](https://groups.google.com/forum/#!forum/mrjob) lately from people who are trying to use mrjob with non line-based data and finding it a struggle.

Here's what I propose. Rather than `read()` and `write()`, protocols will need to define `decode()` and `encode()` as follows:

``` python
def decode(self, input):
    """Read bytestrings from *input* (a generator) and yield tuples of (key, value)."""
    ...

def encode(self, output):
    """Read tuples of (key, value) from *output* (a generator) and yield bytestrings."""
    ...
```

The `--strict-protocols` flag doesn't matter to these new-style protocols (they're always strict). If you really needed loose protocols, you could definitely build a protocol that could detect bad data, increment a counter, and continue (you'd have to pass the job or the `increment_counter` method to the protocol's constructor). Kind of depends how your data is divided into records; if it's line-based, you just discard bad lines, but if you were reading, say, XML, an open tag in the wrong place could confuse everything.

We could eventually deprecate the old-style protocols, though hiding them in an obscure corner of the documentation would probably be sufficient to reduce confusion. I mean, there's nothing wrong with the old interface; I'd just rather people only had to know about two methods rather than four.

davidmarin:
The more I read about this, the more I'm realizing handling non-text data with Hadoop streaming is largely undocumented voodoo.

Manifest files (#754) are a more promising option.

davidmarin:
It doesn't look like there's a good way to get Hadoop to just cat your task the file contents. Hadoop _really_ wants to give your job key value pairs. Even if you set `HADOOP_INPUT_FORMAT` to `org.apache.hadoop.mapred.FixedLengthInputFormat`, which isn't totally correct but appears to be the only built-in input format that yields `BytesWriteable`, Hadoop streaming passes you record numbers.

Closing this in favor of manifest files (see #754).

davidmarin:
See b6bc6049811cee1791315d24429787cb6c3195f8 for an implementation of the above (look in `mrjob/examples/mr_phone_to_url.py`).

davidmarin:
And for binary data that Hadoop recognizes (e.g. sequence files), use a `mapper_init()` method to read directly from `self.stdin`. See #1698.

PR #1052, opened 2015-06-24
chunks to lines

davidmarin:
In anticipation of supporting non-line-based protocols (see #715), mrjob is undergoing a transition from generators that yield lines to generators that can yield data in arbitrary chunks of bytes. When you want lines, you can wrap said chunk iterators through `mrjob.util.to_lines()`

(`to_lines()` was formerly named `buffer_iterator_to_line_iterator()`, which is more clear, but awkward to use as a wrapper function. The old name will still work until v0.6.0.)

`to_lines()` used to add a trailing newline if the last chunk didn't have one; it now leaves the data unmolested (see #819).

This also lets `mrjob.util.gunzip_stream()` just yield chunks of data rather than lines, as promised in its docstring.

I updated the implementation `to_lines()` to try to be more efficient: instead of appending to a `bytes`, we maintain a list of chunks, and `join()` them when we're ready to yield a line. Using `tests.test_util.ReadFileTestCase.test_large_bz2_file()` as a benchmark, I determined that... the old and new implementations are just about equally fast. I still believe it's probably a bit better for the cases I care about, but can't prove it yet.

### DIFF
--- a/mrjob/util.py
+++ b/mrjob/util.py
@@ -63,6 +63,16 @@ def bash_wrap(cmd_str):
     return "bash -c '%s'" % cmd_str.replace("'", "'\\''")
 
 
+def buffer_iterator_to_line_iterator(chunks):
+    """Alias for :py:func:`to_lines()`. Will be removed in v0.6.0.
+
+    .. deprecated:: 0.5.0
+    """
+    log.warning('buffer_iterator_to_line_iterator() has been renamed to'
+                ' to_lines(). This alias will be removed in v0.6.0')
+    return to_lines(chunks)
+
+
 def bunzip2_stream(fileobj, bufsize=1024):
     """Decompress gzipped data on the fly.
 
@@ -685,9 +695,6 @@ def to_lines(chunks):
 
     if leftovers:
         yield b''.join(leftovers)
-
-#: Deprecated alias for :py:func:`to_lines`, will be removed in v0.6.0
-buffer_iterator_to_line_iterator = to_lines
 
 
 def unarchive(archive_path, dest):
--- a/mrjob/util.py
+++ b/mrjob/util.py
@@ -63,7 +63,7 @@ def bash_wrap(cmd_str):
     return "bash -c '%s'" % cmd_str.replace("'", "'\\''")
 
 
-def buffer_iterator_to_line_iterator(iterator):
+def yield_lines(iterator):
     """boto's file iterator splits by buffer size instead of by newline. This
     wrapper puts them back into lines.
     """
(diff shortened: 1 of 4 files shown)


mrjob/util.py
```diff
@@ -739,3 +684,44 @@ def unarchive(archive_path, dest):
...
+            else:
+                yield chunk[start:end]
+
+            start = end
+
+    if leftovers:
+        yield b''.join(leftovers)
+
+#: Deprecated alias for :py:func:`yield_lines`, will be removed in v0.6.0
+buffer_iterator_to_line_iterator = yield_lines
```
jcnnghm: Should calls to `buffer_iterator_to_line_iterator` warn?

tests/test_util.py
```diff
@@ -48,34 +49,49 @@ class BufferIteratorToLineIteratorTestCase(TestCase):
...
+            list(yield_lines(_ for _ in ())),
             [])
 
     def test_buffered_lines(self):
         self.assertEqual(
-            list(buffer_iterator_to_line_iterator(chunk for chunk in
-                                                  [b'The quick\nbrown fox\nju',
-                                                   b'mped over\nthe lazy\ndog',
-                                                   b's.\n'])),
+            list(yield_lines(chunk for chunk in
```
jcnnghm: I think it's more important to maintain a consistent style, but we've been formatting new code like:

```
self.assertEqual(
    list(
        yield_lines(
            chunk for chunk in
            [
                b'The quick\nbrown fox\nju',
                b'mped over\nthe lazy\ndog', 
                b's.\n'
            ]
        )
    ),
    [b'The quick\n', b'brown fox\n', b'jumped over\n', b'the lazy\n', b'dogs.\n']
)
```
Feel free to keep formatting as-is, I don't feel strongly about it either way.

jcnnghm:
Code looks good to me, no blockers.  Feel free to merge.

mrjob/util.py
davidmarin: Good point. Reworked this so that a warning is issued.

tests/test_util.py
davidmarin: Cool. mrjob uses PEP8 for style (see the [release checklist](https://github.com/Yelp/mrjob/wiki/Release-Checklist).

Does Yelp have a maximum line length?

davidmarin:
Okay, cool. After sleeping on it, decided `to_lines()` was a better (mostly, shorter) name than `yield_lines()`.

Other than making `buffer_iterator_to_line_iterator()` warn, the other changes are just renames and documentation, so planning to merge if tests pass.

jcnnghm:
Sounds good to me.

Issue #982, opened 2015-03-17
self._job_name is confusing


davidmarin:
There's an instance variable self._job_name, which is meant to be unique for each job. We use it, for example, to pick the name of the job's temp directory.

This seems like it ought to be the same as the job name in Hadoop (`mapreduce.job.name`), but it's not. This led to a ticket to set `self._job_name` directly (#955), and my thinking that the job could see the value of `self._job_name` by reading the jobconf variable from the environment (#915).

Bottom line, `self._job_name` is just a confusing name. Something like `self._job_key` would avoid that confusion and point out that it's supposed to be unique.

davidmarin:
Apparently there's an option that lets you set job_name, which is probably a bad idea.

PR #1053, opened 2015-06-25
"job key", not "job name"

davidmarin:
mrjob selects a unique key for each job, which is used for selecting temp directories and "locking" for job flow pooling. Because it contains script and user name, it can also be used to audit usage.

This changes the name of the unique key from "job name" (which is easily confused with Hadoop's job names) to "job key". It also removes code which appeared to allow you to set this key manually, but actually does nothing.

See #982 for why this matters.

### DIFF
--- a/docs/guides/configs-hadoopy-runners.rst
+++ b/docs/guides/configs-hadoopy-runners.rst
@@ -68,7 +68,7 @@ Options available to hadoop and emr runners
     :set: all
     :default: script's module name, or ``no_script``
 
-    Description of this job to use as the part of its name.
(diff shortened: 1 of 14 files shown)


smallredbean:
I'm cool with the change. Ship it!

Need a minor version bump, since "--job-name" is removed, right?

davidmarin:
Right, this is for v0.5.0. Thanks!

smallredbean:
SHIP :)

davidmarin:
Also, when I traced through the code, I found `--job-name` didn't actually do anything. :)

PR #1058, opened 2015-06-26
Add support for applying metadata tags to the EMR cluster right after its creation

alejandro-rivera:
This has been asked in https://groups.google.com/forum/#!msg/mrjob/bLGuo4rssds/eWZfKZrSTKQJ.

The third comment in the thread suggests using a **bootstrap action** script in order to have the tags applied. However, I found that solution too cumbersome for something that should be simple to achieve.

Maybe this feature can make it into the 0.5.0 release?

Cheers!


davidmarin:
Wow, that's the most polished pull request I've gotten in a long time! Sure, it's in.

Issue #1028, opened 2015-05-24
connect to buckets through their own region's endpoint


davidmarin:
If you want to write to a bucket that's in, say `us-west-2`, boto will let you connect to it on the default endpoint and write to it. However, this can sometimes result in a `broken pipe` socket error.

What's going on is that S3 issues a 307, redirecting you to the correct endpoint. Somehow, this doesn't work out so well. For more details, see [this boto issue](https://github.com/boto/boto/issues/621).

What we're supposed to do is connect to buckets on the endpoint corresponding to their region.

`mrjob` has largely avoided this issue because it was created at a time when bucket regions *had* to match EMR regions, and issues a warning if you try to cross regions (see #1027). It turns out that nowadays, crossing regions is fine; you just have to connect to the right endpoint.

This turns out to be a fairly invasive change; the current code basically assumes that each runner is only going to need one S3 connection. It doesn't cache it, exactly, just tries to reuse it aggressively.

In most cases (all but getting the list of buckets, I think), we're going to want to pass around buckets rather than connections.

davidmarin:
Might be able to fix #1057 at the same time.

davidmarin:
What to do with the `s3_endpoint` option? Probably it should remain as a workaround for hostile environments, but we should recommend against using it.

davidmarin:
Going to avoid using `boto.s3.connect_to_region()` in favor of just setting `host`. Historically we haven't like to rely on boto to update their region info (that's why we have `mrjob.aws`).

Issue #1027, opened 2015-05-24
"location constraints" are more like guidelines


davidmarin:
It used to be that if you wanted to use EMR in a region other than `us-east-1`, S3 buckets that it used had to have a "location constraint" putting them in the same region. Thus if you try to cross regions, mrjob issues a warning like this:

```
$ python examples/mr_wc.py -r emr README.rst --s3-scratch-uri s3://davidmarin-us-west-2/tmp --aws-region us-east-1
warning: aws_region (us-east-1) does not match bucket region (us-west-2). Your EC2 instances may not be able to reach your S3 buckets.
```

Nowadays, buckets are hosted in particular regions, but EMR in any region can access a bucket in any region. So we don't need this warning.

davidmarin:
This warning is incorrect, but mrjob doesn't exactly work correctly across regions (see #1028).

For now, it's better to have a warning, even if it says the wrong thing, than to later get a broken pipe error with no explanation. Moving to v0.5.0.

davidmarin:
Buckets still have to be created with the correct location, which the API still refers to as a "location constraint". Also, the correct "location constraint" for `us-east-1` is still `''`, not `'us-east-1'`.

Issue #1025, opened 2015-05-24
default region to us-west-2


davidmarin:
Currently, mrjob launches jobs in `us-east-1` by default.

`us-west-2` is probably more correct. [According to the docs](http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-plan-region.html), if you launch an EMR job through the web interface on an account created on or after October 12, *2012*.

`us-west-2` also happens to be one of two [carbon-neutral regions](http://aws.amazon.com/about-aws/sustainable-energy/) available to general users.

This is pretty important not to screw up; basically, if you've been using mrjob already, and haven't set region in your config, it should continue to work when you upgrade to v0.5.0. Part of that will be fixing #687; should also do some extensive hand-testing.

davidmarin:
The nice thing is that EMR can now connect to S3 across regions, so this is harder to screw up. More important than #687 is #1027.

davidmarin:
The main hiccup this will cause for users is that EC2 key pairs exist on a per-region basis. If someone is using a key pair (e.g. to SSH tunnel to the job tracker) but hasn't set `aws_region`, they're either going to need to set `aws_region` to `us-east-1` or create a new key pair for `us-west-2` and update their config to point to it.

Issue #687, opened 2013-08-04
check location constraint on auto-created buckets


davidmarin:
If you don't set `s3_scratch_uri`, `EMRJobRunner` will create a bucket `mrjob-<random hex>` for you to use as scratch space. To prevent a proliferation of auto-created buckets, `EMRJobRunner` first looks for suitable buckets belong to your account named `mrjob-<some hex>`.

So far so good. Where it gets wonky is when you set `aws_region`

Here's what happened to me:

* I ran mrjob with `aws_region` and `s3_scratch_uri` unset. It reused a regionless bucket `mrjob-9...` that it had created previously.

* I ran mrjob with `aws_region` set to `us-west-1`. mrjob knows that my regionless bucket won't work with `us-west-1`, so it created a new bucket `mrjob-7...` with the `us-west-1` region constraint. So far so good.

* I decided to go back to the original settings, and was surprised to find that somehow, I was still running in `us-west-1` anyway. What happened?

So since `aws_region` wasn't set, mrjob essentially assumed that any region would be okay. It used the `mrjob-<hex>` bucket which came first alphabetically, which, due to random chance, was the `us-west-1` one.

This is bad behavior; it's both counterintuitive and non-deterministic.

What mrjob should have done is picked the `mrjob-<hex>` bucket that had the same location constraint (or lack thereof) as it would have selected if it were to create a new `mrjob-<hex>` bucket.

davidmarin:
Since it's now okay to run in a different region from our bucket (see #1027), we shouldn't be inferring region from scratch bucket at all.

PR #1063, opened 2015-06-29
S3 and AWS region changes

davidmarin:
mrjob was built a long time ago, when, in general, if you ran an EMR job in a particular region, it had to use S3 buckets in the same region. (There was an exception for reading from so-called "regionless" buckets in the `us-east-1` region.)

So the mrjob code tried really hard to make sure you were running jobs in a region that would work. If you didn't set `aws_region`, it would actually try to infer region from your scratch bucket. (This led to some odd behavior when the scratch bucket was automatically selected by mrjob; see #687.)

Nowadays, EMR in any region can connect to S3 buckets in any region (see #1027). So some simpler logic is in order:
 * if you don't set `aws_region`, it defaults to `us-west-2` (see #1025 for why)
 * region is never inferred from scratch bucket because it doesn't *have* to match
 * if you don't set `s3_scratch_uri` explicitly, mrjob ensures that the scratch bucket it auto-selects/creates matches your EMR region (because it's more efficient), fixing #687 
 * rather than choosing S3 endpoint based on `aws_region`, we choose it for each bucket use based on its location constraint (see #1028).

The `s3_endpoint` option is actually a little dangerous now, because if you choose a region-specific endpoint, the S3 API won't let you connect to buckets in other regions. Made explicit that `emr_endpoint` and `s3_endpoint` are mostly meant to provide workarounds and don't normally need to be used.

This also removes a lot of code that only existed to support very old versions of boto (see #980).


mrjob/aws.py
```diff
@@ -158,7 +158,7 @@
 _S3_REGIONLESS_ENDPOINT = 's3.amazonaws.com'
 
 # us-east-1 doesn't have its own endpoint or need bucket location constraints
```
jcnnghm: Didn't the bucket location constraints go away?  If so, this comment and constant name should be reworked.

mrjob/options.py
```diff
@@ -288,11 +288,11 @@ def add_emr_opts(opt_group):
         opt_group.add_option(
             '--aws-availability-zone', dest='aws_availability_zone',
             default=None,
-            help='Availability zone to run the job flow on'),
+            help='Availability zone to run EMR jobs in.'),
 
         opt_group.add_option(
             '--aws-region', dest='aws_region', default=None,
```
jcnnghm: Why not set the default here instead of in the code.

mrjob/options.py
```diff
@@ -526,9 +526,9 @@ def add_emr_opts(opt_group):
 
         opt_group.add_option(
             '--s3-endpoint', dest='s3_endpoint', default=None,
-            help=('Host to connect to when communicating with S3 (e.g.'
-                  ' s3-us-west-1.amazonaws.com). Default is to infer this from'
-                  ' region (see --aws-region).')),
+            help=('Force mrjob to connect to S3 on this endpoint (e.g.'
```
jcnnghm: Maybe this option should be dropped, or there should be a note here encouraging people not to use this, since mrjob connects to the right region automatically now.

mrjob/options.py
davidmarin: We almost never set defaults in command-line options, since there's already a system for defaulting runner options (which can be set from either the command line or config file).

(Also there *isn't* a default for `aws_availability_zone`, default is to let the EMR API decide.)

mrjob/options.py
davidmarin: - [ ] make help for `--s3-endpoint` more explicit that you shouldn't use it

davidmarin: Don't want to drop this option because I want to give people with a weird network or in a weird corner of AWS (e.g. GovCloud) a possible workaround.

jcnnghm:
:ship: if test failures are unrelated.

davidmarin:
Thanks!

Yeah, totally unrelated, that's a sporadic (and very minor) test failure that only happens on PyPy. Just pushed a patch for it to `master`.

PR #1068, opened 2015-06-30
update IAM code to use "new" boto methods

### LINKED ISSUE #1062: use IAM API methods in boto
boto 2.35.0 now has methods for (some of) the API calls we use. Where the methods exist, we should use them.

Should also clean out unused methods from `tests.mockboto`.

davidmarin:
Now that we're on boto 2.35.0, boto's `IAMConnection` has methods for most of the IAM API calls we need. This pull request uses those methods instead of setting up calls manually with `IAMConnection.get_response()`. (Fixes #1062)

This also removes a lot of test code for inline (as opposed to "attached") role policies, which we haven't used since v0.4.5.

### DIFF
--- a/mrjob/iam.py
+++ b/mrjob/iam.py
@@ -105,31 +105,22 @@ def _unwrap_response(resp):
     raise ValueError(resp)
 
 
-def _get_response(conn, action, params, path='/', parent=None,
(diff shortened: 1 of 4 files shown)


jcnnghm:
:shipit:

PR #1071, opened 2015-07-01
Write as binary mode when outputting setup-wrapper.sh.

eegeo-devops:
This is so setup-wrapper.sh has correct UNIX line endings for AWS EMR. 
Python will write \r\n on Windows when encountering a \n when writing using text mode, which causes jobs to not start on EMR.

### DIFF
--- a/mrjob/runner.py
+++ b/mrjob/runner.py
@@ -917,9 +917,9 @@ class MRJobRunner(object):
         for line in contents:
             log.debug('WRAPPER: ' + line.rstrip('\n'))
 
-        with open(path, 'w') as f:
+        with open(path, 'wb') as f:
             for line in contents:
-                f.write(line)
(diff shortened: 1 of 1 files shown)


davidmarin:
Interesting! Glad to hear mrjob still mostly works on Windows. :)

This isn't quite the right solution (the lines are bytes on Python 2, and the only reason you can call `.encode()` on them is because they're ASCII), but it won't take me too long to fix this correctly, now that you've alerted me to the problem. Thanks!

davidmarin:
Filed a ticket on it, and started work in my `script-newlines` branch. I merged your change, so your name will appear in the commit logs even if I end up overwriting your changes. :)

marksimpson82:
Thanks! All seems to be working fine :+1:

davidmarin:
Glad I could help.

Issue #1089, opened 2015-07-21
alphabetize switches in --help


davidmarin:
There's probably some simple way to alphabetize --help by long switch name or whatever. We should do this.

davidmarin:
I think it's just `option_parser.option_list.sort(...)`. The options have a `dest` attribute exposed, but `_long_opts` and `_short_opts` are "hidden", so should be careful about using these.

davidmarin:
Want to hold off on this until v0.5.0, so I can test against Python 3.

Issue #1087, opened 2015-07-20
create-job-flow lacks --iam-instance-profile switch


davidmarin:
Somehow the `--iam-instance-profile` switch isn't getting pushed through to `mrjob create-job-flow`.

See [this Google Group thread](https://groups.google.com/forum/#!topic/mrjob/Aurb85zV2s4).

davidmarin:
Yeah, looks like we're using a hard-coded list of options rather than the `mrjob.options.add_*_opts()` functions.

davidmarin:
Turns out support for switches was pretty spotty throughout tools; the assumption was that you'd use a mrjob.conf file. Fixed and future-proofed as best I could.

PR #1090, opened 2015-07-21
Add missing tool switches

davidmarin:
Someone pointed out on in the Google Group that one of the new IAM options wasn't available in the `create-job-flow` tool (see #1087). Looked over the tools, and found that a lot of options were missing; I think we were mostly expecting people to load options from conf files.

This should add all relevant options to each tool in a way that's somewhat future-proof.

I also stopped using option groups in the tools, and alphabetized (see #1089) command-line switches by their destination throughout (including for mrjobs themselves).

Also added the `mrboss` tool to the `mrjob` command line tool as `mrjob boss`. Deprecated the `job_flow_pool` tool because it's never been part of the `mrjob` command line tool and doesn't seem to do anything particularly useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 .tox
 .coverage
 *,cover
+j-*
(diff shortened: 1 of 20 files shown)


bchess:
Ship it!

PR #1094, opened 2015-07-23
RetryWrapper raises last exception if all the tries fail

sudarshang:


davidmarin:
What if `__max_tries` is 0?

sudarshang:
good catch! Will fix and resubmit.

davidmarin:
Thanks!

sudarshang:
updated the PR.

davidmarin:
Merged this into v0.4.x too.

Issue #1115, opened 2015-08-22
could open ssh tunnel earlier


davidmarin:
If we're connected to a cluster with a job already running, we should still be able to get at the job tracker/resource manager before our job starts running.

kkazlowski:
I don't know if this is related, but I have a job that seems to remain in the "PENDING" state (per `step.status.state`, whereas the cluster has gone from "STARTING" to "BOOTSTRAPPING" to "RUNNING" (per `cluster.status.state`).

Found this via our logging in emr.py, around line 1550. 

Due to the `step.status.state` not being "RUNNING", we're not opening the ssh tunnel to the job tracker.

davidmarin:
Yeah, that's pretty much the issue.

Does your step eventually enter the `RUNNING` state? If not, that's a bigger problem that not being able to reach the resource tracker.

kkazlowski:
One of my jobs just seems to be stuck in `PENDING`. Though another makes it through just fine, and I do get access to the resource manager for that one. They're configured identically in terms of mrjob.conf, just different python jobs. Though I know for a fact that both work just fine on mrjob-0.4.4.

Issue #1114, opened 2015-08-22
break up _wait_for_job_to_complete()


davidmarin:
This is in `mrjob.emr`. Like the TODO comment says, this is just too big to write tests for.

Issue #1113, opened 2015-08-21
mrjob ssh subcommand


davidmarin:
After a day of struggling with SSH tunneling, I think it'd be helpful to have a `mrjob ssh` command. It could be more useful than `aws emr ssh` in the following ways:

 * read mrjob.conf
 * allow dynamic tunneling (see #1112)
   * output a config to use with FoxyProxy
 * set up tunnel to job tracker/resource manager, so that it doesn't go up and down and change port with each job
 * create a fake known-hosts file to get around key changing

Issue #1112, opened 2015-08-21
support dynamic SSH tunneling


davidmarin:
Would be nice to be able to have mrjob create a dynamic SSH tunnel, as described in #1013, and have documentation to support it.

PR #1116, opened 2015-08-24
SSH to resource tracker on 3.x AMIs

davidmarin:
This get the SSH tunneling behavior that we know and love working on the 3.x AMIs, fixing #1013. These AMIs use Hadoop 2 (a.k.a. YARN), which has a "resource manager" instead of a "job tracker". You can see job progress and download logs from the resource manager, but not get counters. 

We now opening up the tunnel to the correct port, show the correct URL to point the browser at (it has a different path), use "resource manager" or "job tracker" as appropriate in logging messages, and parse the HTML/JS accessible via the tunnel to show job progress. Also renamed `ssh_tunnel_to_job_tracker` to the shorter and more correct `ssh_tunnel` (the old option/switches still work, with a deprecation warning).

EMR has a tantalizing but ultimately unusable proxy to one of the "application servers" for a job (which do provide counters); see #1013 for discussion.

I found a *lot* of things I'd like to improve about SSH tunneling (see #1112, #1113, #1114, #1115), but for v0.5.0, I'm trying to focus on just getting mrjob's existing features to work with the newer AMIs.

There are tests for the parsers of job progress, but I mostly relied on extensive hand-testing for this change. `_wait_for_job_to_complete()` really needs to be broken up/rewritten (see #1114) before it's possible to write good unit tests for stuff like this, and that would delay the v0.5.0 release.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -20,6 +20,8 @@ v0.5.0, 2015-??-?? -- ???
        * connect to each S3 bucket on appropriate endpoint (#1028)
        * create/select temp bucket in same region as EMR jobs (#687)
      * added iam_endpoint option (#1067)
+     * SSH tunnel now works on 3.x AMIs (#1013)
+       * renamed ssh_tunnel_to_job_tracker option to ssh_tunnel
      * removed s3_conn args from methods in EMRJobRunner and S3Filesystem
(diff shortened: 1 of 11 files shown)


zachm:
Looks great! Nice job @davidmarin :)

Issue #1117, opened 2015-09-01
steps subdirs on EMR are now step IDs


davidmarin:
Noticed this while fixing #888.

When we fetch logs, we currently look inside `steps/<step_num>/`. But EMR now uses step IDs (which look like `s-AAAAAAAA`), so we need to first look up the mapping between step num and step ID with `DescribeSteps`, and then look in the corresponding directory (`steps/s-AAAAAAAA/`).

This is true both on S3 and via SSH, and true on both 2.x and 3.x AMIs.

Issue #888, opened 2014-03-19
MRJob fails to find logs when errors occur on EMR with AMI 3.0.4


boosh:
When running a job that fails on EMR with AMI 3.0.4, mrjob raises an IOError:

```
find: `/mnt/var/log/hadoop/history/': No such file or directory
```

It should look in another subdirectory of `/mnt/var/log/hadoop`, probably `userlogs`.

davidmarin:
Here's the documentation for log files on EMR:

    http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-manage-view-web-log-files.html

okomestudio:
Just reporting that the issue happens with AMI versions 3.7.0 and 3.8.0 as well (did not test with versions earlier than those).

davidmarin:
Thanks! Really truly planning to fix this for v0.5.0.

Smerity:
Until the fix lands in v0.5.0, does anyone know what the most recent / recommended AMI for running jobs with would be?
(all the comments above show is that 3.0.x, 3.7.0, and 3.8.0 are problematic)

davidmarin:
Right now, just looking at fetching counters.

On Hadoop 1 (tried this on AMI 2.4.9), counters are in a path like this:

```
/mnt/var/log/hadoop/history/done/version-1/ip-172-31-29-201.us-west-2.compute.internal_1441062912502_/2015/08/31/000000/job_201508312315_0001_1441062985499_hadoop_streamjob1474198573915234945.jar
```

and have a format like this:

```
MapAttempt TASK_TYPE="SETUP" TASKID="task_201508312315_0001_m_000005" TASK_ATTEMPT_ID="attempt_201508312315_0001_m_000005_0" START_TIME="1441062989982" TRACKER_NAME="tracker_172\.31\.29\.201:localhost/127\.0\.0\.1:55266" HTTP_PORT="9103" .
[...]

Issue #465, opened 2012-05-30
Distinguish between mapper and reducer counters when parsing/printing


irskep:
We currently read the aggregate COUNTER field of a job's output. We should also dump MAP_COUNTERS and REDUCE_COUNTERS separately.

davidmarin:
This is actually probably a bad idea. It looks like we won't be able to get counters separated by map/reduce when we fetch them from S3 when using 3.x AMIs is; see [this comment](https://github.com/Yelp/mrjob/issues/888#issuecomment-136559209).

So, resolving this issue by deciding not to do it. Success!

davidmarin:
It looks like the YARN history file (which is normally stored on HDFS!) has separate map/reduce counters, if we ever want to do this.

PR #1118, opened 2015-09-01
Fetching counters via SSH from 3.x AMIs

davidmarin:
This is part of very slow progress towards getting log fetching/parsing working on the "new" (3.x) AMIs (see #888). This specifically makes it possible to fetch and parse counters via SSH. Hand-tested this, and added some tests for the regexes involved. Basically it boils down to a different filename, and a new, JSON-based format for counters.

Apologies for moving so slowly; the log parsing code is not super good (@irskep, the original author, confirms this ;) ) and I hadn't really taken the time to familiarize myself with it until now.

Also, to make matters worse, mrjob currently can't fetch step logs at all, thanks to EMR now using step IDs rather than step numbers in log paths (see #1117). Will fix that next.

### DIFF
--- a/mrjob/logparsers.py
+++ b/mrjob/logparsers.py
@@ -55,6 +55,7 @@ EMR_JOB_LOG_URI_RE = re.compile(
     r'[_-]hadoop[_-]streamjob(\d+).jar'
     r'(-[A-Za-z0-9-]+\.jhist)?' # this happens on YARN
     r'$')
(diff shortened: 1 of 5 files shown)


davidmarin:
On the plus side, messing with this stuff convinced me to drop #465 (breaking out counters by map/reduce).

bchess:
ship it!

PR #1122, opened 2015-09-04
handle step IDs in step log paths

### LINKED ISSUE #1117: steps subdirs on EMR are now step IDs

davidmarin:
This fixes #1117.

In brief: log paths for step logs used to use numbers (`1`, `2`, etc.) but now use randomly-assigned step IDs (`s-AAAAAAAA`). This means we need an extra mapping, both to ignore steps that don't matter for our job, and to sort steps in the proper order so that we can find the earliest failure.

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -1750,7 +1750,11 @@ class EMRJobRunner(MRJobRunner):
 
     def ls_step_logs_ssh(self, step_nums, cluster_step_ids):
         # step nums are 1-indexed
-        step_ids = [cluster_step_ids[step_num - 1] for step_num in step_nums]
+        if step_nums is None:
+            step_ids = None
+        else:
+            step_ids = [cluster_step_ids[step_num - 1]
+                        for step_num in step_nums]
 
         self._enable_slave_ssh_access()
         return self._enforce_path_regexp(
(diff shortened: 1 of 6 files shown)


mrjob/emr.py
```diff
@@ -1680,18 +1680,26 @@ def _step_output_uri(self, step_num):
...
+    def _enforce_path_regexp(self, paths, regexp, filters=None):
         """Helper for log fetching functions to filter out unwanted
-        logs. Only pass ``step_nums`` if ``regexp`` has a ``step_nums`` group.
+        logs.
+
+        filters maps from regexp group to a list of thing to match
         """
+        filters = dict((group_name, set(str(v) for v in values))
+                       for group_name, values in (filters or {}).items()
+                       if values)
```
bchess: Nested list comprehensions are kinda hard to read. Break it out?

bchess:
Fix 'n ship

davidmarin:
Fixed, shipped

Issue #1123, opened 2015-09-04
bring sanity to log handling code


davidmarin:
Being able to tell what might have caused your job to fail is a really important part of mrjob. Unfortunately, the code that backs it is really challenging to maintain.

Proposed way to organize the refactored code:

 * `mrjob.logs.ls`: Identifying and sorting log paths. Currently spread between `mrjob.emr` and `mrjob.logparsers` (which should be deleted).
 * `mrjob.fs`: Code to actually download logs (mostly already there, but like, `_enable_slave_ssh_access()` should happen automatically)
 * `mrjob.logs.parse`: finding failures in logs (currently mostly in `mrjob.parse`)

davidmarin:
Also, there's a lot of fumbling to figure out the address of the master node, log dir, step IDs, etc. `mrjob` should just cache this information as it learns it.

davidmarin:
Okay, this is starting to come together.

Basically there are three kinds of logs worth parsing:
 * task stderr and syslog (in `userlogs/`), to find hadoop/task errors
 * history files (in `history`), to find counters and hadoop errors
 * hadoop command stderr (also in `steps/` on S3), for counters (YARN only) and tracking URL

To get counters, we should parse Hadoop command stderr, and fall back to the history file

To get probable cause of failure, we should try parsing both task stderr/syslog and history. Generally we want the latest error, but we'd rather show task stderr if we can (e.g. if we got a timeout and a Python exception on the same step, we'd rather show the Python exception).

Would be helpful to have a standard error format, probably with these fields:
 * java_error: java stacktrace from Hadoop
   * path: URI of log file where error found
   * line: line in that file where error found
   * error: the error itself
 * task_error: Python exception, etc.
   * path
   * line
   * error
   * command: the last command run (e.g. to catch setup script errors)
 * task_attempt_id (pre-YARN)
 * job_id
 * application_id (YARN)
 * container_id (YARN)

Given this common format, it should be pretty easy to pick the best error (or find all errors).

davidmarin:
Also could add a `bootstrap_error` field for bootstrapping errors on EMR (similar to task_error).

PR #1128, opened 2015-09-16
refactor finding of relevant logs

### LINKED ISSUE #1121: task attempt fetching on 3.x AMIs
(This fixes part of #888)

The path for task attempts in the 3.x AMIs looks very different than on 2.x. An example:

```
s3://mrjob-35cdec11663cb1cb/tmp/logs/j-21QKHYM5WJJHS/task-attempts/application_1441057410014_0001/container_1441057410014_0001_01_000004/stderr.gz
```

The "application" directory appears to be redundant; we can get by with just the "container" part. Not entirely sure what all the numbers mean (e.g. which is the step number? Can we tell which is mapper or reducer?)

davidmarin:
This branch refactors some very old, crufty code for finding log files of various types (so we can, for example, determine why a job failed). A bunch of stuff that used to be handled by `mrjob.emr` and `mrjob.logparsers` is now in `mrjob.ls`. (This partially implements #1123; next step is parsing.) `mrjob.emr` and `mrjob.hadoop` now use the same code (in `mrjob.logs.ls`) to fetch logs.

This also updates the SSH Filesystem (`mrjob.fs.ssh`) so it needs a lot less hand-holding from `mrjob.emr` to get at logs that are only available by tunneling through the master node.

This change also allows `mrjob` can now successfully fetch task and step logs from the 3.x AMIs (this fixes #1121). It can fetch counters via SSH but not S3; still need to figure out which logs they appear in. Essentially, there is some hand-testing in my future before I'm confident this all works on the 3.x AMIs, but I'm submitting the pull request now before diving into fixing up the parsing code so that it'll be of a manageable size.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -29,6 +29,7 @@ v0.5.0, 2015-??-?? -- ???
      * hdfs_scratch_dir option is now hadoop_tmp_dir (#318)
  * removed mrjob.compat.get_jobconf_value() (use jobconf_from_env())
  * mrjob.util:
+   * random_identifier() moved here from mrjob.aws
    * renamed buffer_iterator_to_line_iterator() to to_lines()
      * to_lines() no longer appends a newline to data (#819)
    * removed extract_dir_for_tar()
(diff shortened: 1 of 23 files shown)


bchess:
This lg2m, ship it!

Issue #1070, opened 2015-07-01
pool_wait_minutes from .mrjob.conf ignored


alanhdu:
In our `.mrjob.conf`, we set `pool_wait_minutes: 10`. But unless we pass in the command-line flag `--pool-wait-minutes 10`, mrjob doesn't bother waiting.

I think it's because the default command line option for `--pool-wait-minutes` is 0, which overwrites the option read from `.mrjob.conf` (other defaults are set to `None`, so https://github.com/Yelp/mrjob/blob/master/mrjob/conf.py#L261 catches that).

davidmarin:
Whoops. Will try to fix that for v0.5.0.

davidmarin:
Going to try to backport this to 0.4.6

PR #1132, opened 2015-09-28
load pool_wait_minutes from mrjob.conf

davidmarin:
Fixes a bug where we were specifying the default value in the option rather than in the runner, so that "default" always overrode mrjob.conf (see #1070)


davidmarin:
This is simple enough that I'm not going to bug someone for a code review.

Issue #1035, opened 2015-06-03
bootstrap_python should pip install ujson


davidmarin:
Now that we have the `bootstrap_python` option, we should use it to `pip install ujson`, so that jobs will run faster by default.

We need the version of `pip` that matches our version of Python (so `pip3` if on Python 3). Also, should double check which AMIs have `pip` installed for Python 2.

davidmarin:
Actually, I'm iffy on this feature. Installing Amazon Linux packages seems safe enough, but I don't want people's jobs to fail if PyPI goes down. Might be safer to `vendor` it and roll it into `mrjob_bootstrap`.

davidmarin:
PyPI [uses a CDN](https://pypi.python.org/mirrors). and the [last article I could find about it going down](https://jacobian.org/writing/when-pypi-goes-down/) is from 2010. So I think it's probably safe to treat PyPI as always up. Users can always turn off `bootstrap_python` in a pinch.

davidmarin:
[This post on StackOverflow](http://stackoverflow.com/questions/28403263/canonical-replacement-for-use-mirrors) mentions reliability issues with the PyPI CDN.

davidmarin:
On Python 2 you just `sudo yum install python-pip` and then `sudo pip install ujson`.

Unfortunately, the Python 3 package on the Amazon Linux AMIs is really minimal. There's no `pip`, so you have to [curl and run get-pip.py](http://pip-python3.readthedocs.org/en/latest/installing.html) (despite the fact that pip is supposed to be pre-installed in Python 3.4+). Then you still can't install `ujson` because there's no Python source, just the binaries. Really, if you want `ujson` (or other libraries you have to compile) with Python 3, you need to first install Python 3 from source.

davidmarin:
Probably worth it to install `ujson` on Python 2, anyway.

PR #1136, opened 2015-09-29
Install ujson on Python 2

davidmarin:
This enables the `bootstrap_python` option by default for all versions of Python. On Python 2, rather than installing Python 2 (which is already installed), it auto-installs `pip` and `ujson` for you (so that JSON serialization will magically be fast). See #1035.

Unfortunately, Amazon's support for Python 3 is really minimal, and installing `pip` is simpler than `ujson`. Left `bootstrap_python` as-is for Python 3, but added extensive documentation about what you need to do to get `pip` and/or `ujson`.

Also disabled `bootstrap_python_packages` on Python 3.


davidmarin:
Raising an exception for `bootstrap_python_packages` is a bit harsh; a warning (and not running the command) should do it.

PR #1147, opened 2015-10-16
mkdir -p in Hadoop 2

### LINKED ISSUE #991: HadoopFilesystem.mkdir() should use -p
#845 patched `HadoopRunner._mkdir_on_hdfs()` to use `-p` on Hadoop 2.0+, but we didn't patch `HadoopFilesystem.mkdir()`, so it's probably still broken on newer versions of Hadoop.

Honestly, I'm not entirely sure why `_mkdir_on_hdfs()` exists; we should just be using `self.fs.mkdir()`.

davidmarin:
This makes `HadoopFilesystem.mkdir()` use `-p` when we're on Hadoop 2 (fixes #991).

This required a bit of re-working; it used to be that `HadoopJobRunner` was responsible for figuring out the version of Hadoop we're using, but `HadoopFilesystem.mkdir()` needed version as well, so moved `get_hadoop_version()` from `HadoopJobRunner` to `HadoopFilesystem`.

Also, got rid of the redundant `_mkdir_on_hdfs()` method, and added `mrjob.compat.uses_yarn()`.

### DIFF
--- a/mrjob/compat.py
+++ b/mrjob/compat.py
@@ -757,6 +757,13 @@ def uses_generic_jobconf(version):
     return version_gte(version, '0.20')
 
 
+def uses_yarn(version):
+    """Basically, is this Hadoop 2? This also handles versions in the
+    zero series (0.23+) where YARN originated."""
+    return (version_gte(version, '2') or
+            version_gte(version, '0.23') and not version_gte(version, '1'))
+
+
 def version_gte(version, cmp_version_str):
     """Return ``True`` if version >= *cmp_version_str*."""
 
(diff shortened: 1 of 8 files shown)


davidmarin:
Also, the test case that was supposed to catch this was tripping over itself by re-creating the same directory several times (which resulted in an excusable error message from mockhadoop, obviating the non-zero return code). Often a bad idea to use a for loop to run several versions of a test.

davidmarin:
Hand-tested this on EMR (with Hadoop 2.4.0).

PR #1163, opened 2015-10-27
only check for running yarn applications when yarn command exists

dallasmarlow:


davidmarin:
Sure, it seems to work the way it is, but it's cleaner and more clear to use `which`.

Going to change this a little bit before checking it in. The double brackets don't exist in Bourne shell, and you don't need them in any case; the exit status of `which` tells you whether it exists.

davidmarin:
Also, you got your operator precedence wrong (`||` binds more tightly than `&&`). Did you hand-test this?

dallasmarlow:
@davidmarin i didn't realize that when the yarn executable isn't in the path that it still works and made this change in preparation to deploying to a mixed env. with the operator precedence concern it's part of the conditional expression and works as intended, but if this change is no-op then it doesn't make much sense.

davidmarin:
Well, I think your change makes it a lot more readable, which matters for maintainability. I'm about done merging it anyway. If it looked wrong to you, it'll probably look wrong to someone else, and this will save them time.

What I meant about operator precedence is:

```
[ -z "$LAST_ACTIVE" ] || ! which hadoop ... || nice hadoop ... || which yarn ... && yarn ...
```

is equivalent to:

```
([ -z "$LAST_ACTIVE" ] || ! which hadoop ... || nice hadoop ... || which yarn ...) && yarn ...
```

that is, if `yarn` fails, the whole expression fails, and the script thinks the cluster is idle.

PR #1170, opened 2015-11-03
Fix two bugs when running mrjob master:

mikekap:
Don't attempt to clean up jobs that never started, even when pooling job flows.
Don't fail when get_location throws - e.g.
    `python -c 'import boto; boto.connect_s3().get_bucket("elasticmapreduce").get_location()'`

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -1054,9 +1054,12 @@ class EMRJobRunner(MRJobRunner):
 
     def _cleanup_job(self):
         # kill the job if we won't be taking down the whole job flow
-        if not (self._cluster_id or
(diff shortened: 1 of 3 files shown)


davidmarin:
Oh, that's interesting, you have permission to read/write the bucket but not query its location?

mikekap:
Just read. This is from a bootstrap step.

I'll fix the tests later today.

davidmarin:
Ah, I see. Yeah, sounds useful to be able to read from the `elasticmapreduce` bucket.

mikekap:
Ping

davidmarin:
Merged. Just wanted to make sure I had tests for both of your changes.

PR #1176, opened 2015-11-06
Skip test if no jar command

davidmarin:
This skips `tests.test_util.ArchiveTestCase.test_unarchive_jar()` if there is no `jar` command.

Brought in `which()` from the `hadoop-prefix` branch.

### DIFF
--- a/mrjob/util.py
+++ b/mrjob/util.py
@@ -24,6 +24,7 @@ import os
 import pipes
 import random
 import shlex
+import shutil
 import sys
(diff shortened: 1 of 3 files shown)


davidmarin:
(Hand-tested this by moving the `jar` command on my system.)

davidmarin:
Ended up reverting this in favor of f3d9f5eb2ce5a545d8fbf0b4f853627ee60f9dcd

Issue #1166, opened 2015-10-29
paths in conf include should be relative to conf


davidmarin:
Currently, if you use `include` in a mrjob conf file with a relative path, it's relative to the current working directory. It *should* be relative to the including conf file's directory.

Also, the list of already-loaded conf files should be real paths (see `os.path.realpath`).

davidmarin:
Fixed by #1179.

PR #1179, opened 2015-11-09
relative conf includes

davidmarin:
If you use a relative path in `include:` in a config file, it's now relative to the real path (after resolving symlinks) of the config file.


davidmarin:
include paths are now relative to real path of conf file (fixes #1166)

Issue #1127, opened 2015-09-15
remove fetch_logs tool


davidmarin:
Currently `mrjob.tools.emr.fetch_logs` does nothing if you don't specify a command-line switch.

By default, it should download logs into a directory with the same name as the cluster, rather than catting them. It should also be able to analyze logs already downloaded to a given directory.

It should be fine to take a single argument that's either a cluster ID or a path, with switches to make this explicit.

davidmarin:
Not sure how to fetch logs when we're supposed to get them from several places (e.g. from slave nodes). Unless we use subdirectories which look like URLs (e.g. 's3://...', 'file://...'). Might work.

davidmarin:
What would we want to fetch logs for? Some possible use cases:

 * all logs on a given EMR cluster
 * all logs for a particular run of a mrjob (using the job key) on EMR
 * all logs for a particular step on a particular cluster
 * all logs for a particular application ID on Hadoop
 * all logs for a particular run of a mrjob on Hadoop? (how would we map this to application IDs?)

davidmarin:
Let's just remove the `fetch_logs` tool entirely.

I think what we really want is to expose a way for a runner to programmatically find relevant logs, cause of failure, etc., and a way to re-create a the runner for a past job so we can access its logs (e.g. by looking up the job key).

This involves a lot of choices about what to expose; I want to get log interpretation working before making these choices.

Issue #565, opened 2012-09-20
Use <output_dir>/_logs in Hadoop


tarnfeld:
The [`counters()`](https://github.com/Yelp/mrjob/blob/master/mrjob/hadoop.py#L468) method returns nothing. By the looks of things, when the job completes and the `_fetch_counters()` method is called, it assumes you're running mrjob on the job tracker node (I think) and the logs directory is on the local disk. When running on a remote hadoop cluster, the local logs/history directory doesn't exist where mrjob is being run and therefore no logs/counters are found.

By the looks of things, the counters can also be found in the `_logs` directory of a successful job.

wayoutmind:
+1

davidmarin:
Will look into this.

tarnfeld:
Just a reminder that we've got patches for this fix further up in the comments (from 2 years ago 😱)

davidmarin:
Sorry to take so long to get to this. I'm partway through revamping the log parsing system, and it should be pretty easy to add once I do that.

I'm more worried about testing; it looks like `_logs` in the output dir is a Cloudera-specific feature. What's the easiest way for me to get Cloudera up and running so I can poke at it?

davidmarin:
Looks like I should try out `[Cloudera Live](http://www.cloudera.com/content/www/en-us/developers/get-started-with-hadoop-tutorial.html)`.

Issue #1195, opened 2015-12-07
Find logs on Hadoop


davidmarin:
Need an approach that can find the relevant logs for at least some versions of Hadoop.

davidmarin:
Some articles about where to find Hadoop logs:

http://hortonworks.com/blog/simplifying-user-logs-management-and-access-in-yarn/

davidmarin:
Okay, it's simpler than that; `$HADOOP_PREFIX/logs` is a reasonable place to look.

The task logs are in `<log dir>/userlogs/<application ID>`, so we need to read application ID from Hadoop's output.

davidmarin:
Looks like `logs.ls` currently has too much EMR-specific stuff in it. We should leave it up to the runner to determine the root directories to look in, and search inside those.

davidmarin:
This is pretty much done. Working on counters/history log now.

PR #1199, opened 2015-12-23
Fetch hadoop logs in YARN

davidmarin:
This change makes the Hadoop runner able to fetch and parse task logs, and determine cause of error when a job failed due to a Python exception. Currently this only works in YARN. Automatically tries all the possible places logs could be hiding. (See #1195). 

Also added a `hadoop_log_dirs` option, in case users want to hard-code the log directory location instead of using environment variables (this mostly exists so we have a place in the docs to explain the automatic log-finding behavior if this is *not* set).

This also cleans up the existing code for parsing counters from the hadoop binary's stderr, moving more stuff into functions and other modules.

This removes `fetch_logs`, which no longer makes sense with the new way of handling logs (see #1127).


davidmarin:
Fetch hadoop logs in YARN. Fixes #1127, fixes #565, addresses #1195

PR #1210, opened 2016-01-18
Job logs on AWS EMR 3.11.0 are compressed.

ramrengaswamy:
AWS EMR 3.11.0 creates compressed job logs.
Therefore, the regexp for job uri should accommodate S3 objects that end in `.gz`


davidmarin:
I'm actually about to delete this code and have EMR use code in https://github.com/Yelp/mrjob/blob/master/mrjob/logs/history.py instead, but... until then, wny not? Thanks for your patch.

PR #1214, opened 2016-01-27
List the 3 and 4 AMIs before the 2 AMI in the docs

alex:
the debian squeeze base used in the 2 AMI is an unsupported debian with an unsupported openssl

### DIFF
--- a/docs/guides/emr-bootstrap-cookbook.rst
+++ b/docs/guides/emr-bootstrap-cookbook.rst
@@ -151,28 +151,6 @@ Installing System Packages
 EMR gives you access to a variety of different Amazon Machine Images, or AMIs
 for short (see :mrjob-opt:`ami_version`).
 
-2.x AMIs
-^^^^^^^^
-
-The 2.x AMIs are based on `Debian 6.0.2 (Squeeze)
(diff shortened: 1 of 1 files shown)


alex:
(The goal of this PR is to encourage people to use teh newer versions, inc ase that was unclear)

davidmarin:
Yeah, okay, people probably shouldn't be using the 2.x AMIs anymore. :)

alex:
@davidmarin It could also be removed if you really want to encourage people to use the latest :-)

davidmarin:
Heh. That would just be mean. If people are already on 2.x, they deserve documentation.

alex:
:+1:

On Thu, Mar 3, 2016 at 2:03 PM, David Marin <notifications@github.com>
wrote:

> Heh. That would just be mean. If people are already on 2.x, they deserve
> documentation.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Yelp/mrjob/pull/1214#issuecomment-191915990>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

Issue #1230, opened 2016-02-26
EC2 key pair not included in pooling criteria


davidmarin:
If the user specifies `ec2_key_pair`, they shouldn't attempt to join a cluster with a different `ec2_key_pair`.

(If `ec2_key_pair` isn't specified, it shouldn't matter what the cluster's EC2 key pair is.)

PR #1239, opened 2016-03-05
Don't join an EMR cluster with a different ec2_key_pair.

mmontagna:
This PR should fix issue #1230. 

Sorry there are no tests for this particular feature, I couldn't quickly figure out how to trigger this behavior in the test suite, but it's been hand tested. 

If this is not useful please ignore.


davidmarin:
Thanks!

Issue #1241, opened 2016-03-08
disable _cleanup_job() in EMR


davidmarin:
This seems to be broken in a number of ways (can't handle empty job list, raises `IOErrors` that are silently swallowed).

This provides a pretty small amount of value (basically, allowing you to stop a potentially successful job on a shared cluster you've launched by hitting ctrl-c), and does it in a messy way (basically, SSHing in and killing any job it can find). Really, we need something more surgical, but for now, we need this not to be broken.

davidmarin:
Should note in the docs that the `JOB` cleanup option currently does nothing.

PR #1243, opened 2016-03-08
Google dataproc runner

davidmarin:
Minimal viable product for Google Cloud Dataproc support.

This pull request was mostly written by @mtai, not me. :)

This version doesn't perform any log interpretation (counters, cause of error). It also doesn't support any sort of pooling, though it's not entirely clear if that matters with Dataproc's billing model and quick startup. And, you know, there might be bugs. Seems pretty well unit-tested, at least!


mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+import json
+import logging
+import os
+import os.path
+import pipes
+import time
+import re
+
+try:
+    from oauth2client.client import GoogleCredentials
```
davidmarin: TODO: add these new libraries to setup.py

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+
+    def __init__(self, alias, opts, conf_paths):
+        super(DataprocRunnerOptionStore, self).__init__(alias, opts, conf_paths)
+
+        # TODO - mtai @ davidmarin - As of 2016-02-26, the ONLY valid region is "global"
+        self['gcp_region'] = _DEFAULT_GCP_REGION
+
+        # Dataproc requires a master and >= 2 worker instances
+        # gce_num_instances refers ONLY to number of WORKER instances and does NOT include the required 1 instance for master
+        # In other words, minimum cluster size is 3 machines, 1 master and "gce_num_instances" workers
```
davidmarin: This is similar enough to EMR that the Dataproc and EMR runners should use the same options and agree on what they mean. So, like, there should just be a `num_instances` option, and it should have to be at least 2 for Dataproc.

It's better to raise an exception than silently fix something like this.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        if self['gce_num_instances'] < 2:
+            self['gce_num_instances'] = 2
+
+    def default_options(self):
+        super_opts = super(DataprocRunnerOptionStore, self).default_options()
+        return combine_dicts(super_opts, {
+            'bootstrap_python': True,
+            'check_api_status_every': 30,
```
davidmarin: This should probably be a shared option too, called `check_cluster_every`.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        # keep track of when we launched our job
+        self._dataproc_job_start = time.time()
+
+    def _wait_for_steps_to_complete(self):
+        """Wait for every step of the job to complete, one by one."""
+
+        total_steps = self._num_steps()
+        # define out steps
+        for step_num in xrange(total_steps):
+            # NOTE - mtai @ davidmarin - Dataproc accepts multiple jobs at once
```
davidmarin: Yeah, these names aren't really appropriate for Dataproc's model. :) You probably want methods named `_run_steps()`, `_launch_step()`, and `_wait_for_step_to_complete()`, or something like that.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+    # TODO - mtai @ davidmarin - Re-enable log support and supporting cleanup
+    def _cleanup_logs(self):
+        super(DataprocJobRunner, self)._cleanup_logs()
+
+    # TODO - mtai @ davidmarin - Revisit
+    def _cleanup_job(self):
+        pass
+
+    def _cleanup_cluster(self):
+        try:
```
davidmarin: You only want to do this if the cluster isn't shared (basically, only for clusters you started).

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        # when did our particular task start?
+        self._dataproc_job_start = None
+
+        # This will be filled by _wait_for_steps_to_complete()
+        self._log_interpretations = []
+
+    @property
+    def dataproc_client(self):
+        if not self._dataproc_client:
+            credentials = GoogleCredentials.get_application_default()
```
davidmarin: Ah, this is how the credentials are added.

You probably at least want a link to https://developers.google.com/identity/protocols/application-default-credentials#howtheywork. And it should be possible to specify the credentials in `mrjob.conf`.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+_HADOOP_STREAMING_JAR_URI = 'file:///usr/lib/hadoop-mapreduce/hadoop-streaming.jar'
+
+# TODO - mtai @ davidmarin - Example invocation options for mrjob.py
+#   --runner dataproc --gcp-project google.com:pm-hackathon --gcp-zone us-central1-b --gcp-cluster vbp04 --no-output gs://boulder-input-data/*.txt --output gs://boulder-input-data/mrjob-boulder
+
+# TODO - mtai @ davidmarin - Re-implement SSH tunneling? - Content pulled from job run output
+# The url to track the job: http://default-taim-m:8088/proxy/application_1456679373146_0002/
+
+# TODO - mtai @ davidmarin - Re-implement logs parsing?  See dataproc metainfo and driver output - ask Dennis Huo for more details
+# 'gs://dataproc-801485be-0997-40e7-84a7-00926031747c-us/google-cloud-dataproc-metainfo/8b76d95e-ebdc-4b81-896d-b2c5009b3560/jobs/mr_most_used_word-taim-20160228-172000-034993---Step-2-of-2/driveroutput'
```
davidmarin: This should be really straightforward; take a look at `mrjob.logs.mixin` and how it's used. Basically, you just have to tell it where to find each kind of log, and it can take it from there.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+                self.fs.rm(self._gcs_tmp_dir)
+                self._gcs_tmp_dir = None
+            except Exception as e:
+                log.exception(e)
+
+    # TODO - mtai @ davidmarin - Re-enable log support and supporting cleanup
+    def _cleanup_logs(self):
+        super(DataprocJobRunner, self)._cleanup_logs()
+
+    # TODO - mtai @ davidmarin - Revisit
```
davidmarin: This is not very important; it basically lets you ctrl-c jobs that you ran by accident without having to kill the cluster. I'm actually temporarily disabling this in EMR because the way it's done is so broken (see #1241).

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        if self._gcs_tmp_dir:
+            try:
+                log.info('Removing all files in %s' % self._gcs_tmp_dir)
+                self.fs.rm(self._gcs_tmp_dir)
+                self._gcs_tmp_dir = None
+            except Exception as e:
+                log.exception(e)
+
+    # TODO - mtai @ davidmarin - Re-enable log support and supporting cleanup
+    def _cleanup_logs(self):
```
davidmarin: The main value this provides is not filling up your cloud storage with logs, which... probably isn't that important.

Honestly, this is kind of an obsolete approach; having a temp directory that expires after 30 days would be better (e.g. https://cloud.google.com/storage/docs/lifecycle).

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+
+        for path, gcs_uri in self._upload_mgr.path_to_uri().items():
+            log.debug('uploading %s -> %s' % (path, gcs_uri))
+
+            # TODO - mtai @ davidmarin - Implement upload function for other FSs
+            self.fs.upload(path, gcs_uri)
+
+    ### Running the job ###
+
+    def cleanup(self, mode=None):
```
davidmarin: Lots of comments about cleanup below; bottom line is that writing code to clean up stuff on cloud storage is a very low priority (you *do* want to make sure to shut down clusters).

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+
+        # make sure we can see the files we copied to GCS
+        log.info('Creating Dataproc Hadoop cluster')
+
+        # NOTE - mtai @ davidmarin - We create the cluster if its missing, otherwise we join an existing one
+        # NOTE - mtai @ davidmarin - The cluster MUST be created before submitting jobs, otherwise job submission fails
+
+        # TODO - mtai @ davidmarin - change the way we auto-create cluster names
+        # "clusterName must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,53}[a-z0-9])?).'"
+        self._gcp_cluster = self._gcp_cluster or "default-%s" % self._opts['owner']
```
davidmarin: Okay, having a single cluster named after `owner` is a terrible hack for a number of reasons. I was expecting the initial implementation would create a cluster named `mrjob-<random hex>`, and shut it down when the job is done (and then we could implemented something like pooling).

The fact the clusters can't shut themselves down (it's all up to `mrjob`) is a bit scary. I think it might be useful to add something like `mrjob/bootstrap/terminate_idle_cluster.sh` to the dataproc runner from the get-go. This script would almost work as is, but we could simplify it a bit: no `sudo`, assume `yarn` command exists, don't worry about how close we are to the end of a billing hour (since Google bills by the minute).

@evilsoapbox -- if you're listening, having Dataproc automatically terminate idle clusters would save your customers a lot of pain and would be a competitive advantage over Elastic MapReduce.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+
+        try:
+            self._get_cluster()
+            log.info('Adding our job to existing cluster %s' % self._gcp_cluster)
```
davidmarin: A couple of things to think about that might inform the design of the Dataproc runner:

Dataproc lets you run multiple jobs on the same cluster; when is this actually useful? (We already have a cloud, so we don't need Hadoop/YARN on a cluster to act as a cloud within a cloud). Some ideas:
 * Latency, maybe? Like, I have a very short job, and I'd rather have it compete with other jobs on an existing cluster than wait for a new cluster to spin up?

In theory, as long as step output goes to GCS and not HDFS, a job could, in theory, run every step on a different cluster, shopping around on a step-by-step basis. Honestly, I can't see much advantage to this; unlike on EMR (where we never tried this), you're going to be paying by the minute for any cluster you use, so there aren't really "available" clusters just waiting around for you to use their idle time.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        for step_num in xrange(total_steps):
+            # NOTE - mtai @ davidmarin - Dataproc accepts multiple jobs at once
+            # NOTE - mtai @ davidmarin - In this particular implementation, we must build each step on the fly and submit it and WAIT for step completion
+            # NOTE - mtai @ davidmarin - we THEN iterate through the loop and continue on
+
+            # Build each step
+            hadoop_job = self._build_dataproc_hadoop_job(step_num)
+
+            # Clean-up step name
+            step_name = '%s - Step %d of %d' % (self._job_key, step_num + 1, self._num_steps())
```
davidmarin: It seems a little ridiculous to create spaces and uppercase letters only to convert them to hyphens and lowercase letters. :)

Was expecting something more like '%s-step-%05d' % (_cleanse_gcp_job_id(self._job_key), step_num)

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        If it fails, attempt to diagnose the error, and raise an
+        exception.
+
+        This also adds an item to self._log_interpretations
+        """
+        log_interpretation = dict(job_id=job_id)
+        self._log_interpretations.append(log_interpretation)
+
+        while True:
+            # don't antagonize Dataproc's throttling
```
davidmarin: What throttling? :)

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+                raise StepFailedException(step_num=step_num, num_steps=num_steps)
+
+    def _step_input_uris(self, step_num):
+        """Get the gs:// URIs for input for the given step."""
+        if step_num == 0:
+            return [self._upload_mgr.uri(path)
+                    for path in self._get_input_paths()]
+        else:
+            # put intermediate data in HDFS
+            return ['hdfs:///tmp/mrjob/%s/step-output/%s/' % (
```
davidmarin: You just copied this from EMR, but realized that we should use `%d` for the step num.

Thinking I'd like to standardize on `'%05d' % step_num` (0-indexed). This is more of a note for myself.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+
+        Create the master bootstrap script and write it into our local
+        temp directory. Set self._master_bootstrap_script_path.
+
+        This will do nothing if there are no bootstrap scripts or commands,
+        or if it has already been called."""
+        if self._master_bootstrap_script_path:
+            return
+
+        # TODO - mtai @ davidmarin - Don't know when we don't need to do this...
```
davidmarin: Definitely need to do this; offhand I can't think of any way that Dataproc differs from EMR here.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        if not self._opts['bootstrap_python']:
+            return []
+
+        # Python 2 is already installed; install pip and ujson
+
+        # (We also install python-pip for bootstrap_python_packages,
+        # but there's no harm in running these commands twice, and
+        # bootstrap_python_packages is deprecated anyway.)
+
+        # NOTE - mtai @ davidmarin - Dataproc uses apt-get, ujson needs python-dev installed
```
davidmarin: Neat. What about Python 3?

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+from mrjob.setup import parse_setup_cmd
+from mrjob.step import StepFailedException
+from mrjob.util import cmd_line
+from mrjob.util import random_identifier
+
+log = logging.getLogger(__name__)
+
+_DEFAULT_GCP_REGION = 'global'
+_HADOOP_STREAMING_JAR_URI = 'file:///usr/lib/hadoop-mapreduce/hadoop-streaming.jar'
+
```
davidmarin: You're going to need a way to specify your machine image, and learn what version of Hadoop it's using.

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        # TODO - very odd to use / notation
+        # gcloud config set core/project google.com:pm-hackathon
+
+        gce_num_instances = int(self._opts['gce_num_instances'])
+        assert gce_num_instances >= 2
+        bootstrap_cmds = [self._upload_mgr.uri(self._master_bootstrap_script_path)]
+
+        # Change machine type
+        zone_uri = 'https://www.googleapis.com/compute/v1/projects/%(project)s/zones/%(zone)s' % dict(project=self._gcp_project, zone=self._gcp_zone)
```
davidmarin: Ugh, there isn't library support for this stuff?

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        self.dataproc_client.clusters().create(
+            projectId=self._gcp_project,
+            region=self._gcp_region,
+            body=cluster_data).execute()
+        cluster_state = None
+
+        # TODO - mtai @ davidmarin - See https://cloud.google.com/dataproc/reference/rest/v1/projects.regions.clusters#State
+        # TODO - mtai @ davidmarin - We should probably pull these strings to constants
+
+        # TODO - mtai @ davidmarin - Can end up in infinite loop if cluster_state doesn't end up in (RUNNING,ERROR)
```
davidmarin: Based on the stat docs URL above, t looks like we'd also want to give up if the state is `DELETING`, and go ahead if the state is `UPDATING`. Other than that, what do we have to worry about?

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        while bool(cluster_state != 'RUNNING'):
+            result_describe =  self.dataproc_client.clusters().get(projectId=self._gcp_project,
+                                               region=self._gcp_region,
+                                               clusterName=self._gcp_cluster).execute()
+    
+            cluster_state = result_describe['status']['state']
+            if cluster_state == 'ERROR':
+                raise Exception(result_describe)
+    
+            log.info('Waiting for cluster to launch... sleeping 5 second(s)')
```
davidmarin: Once everything else is working, let's think about how often we want to print status messages to the user and what they should look like (just cleaned this up in EMR).

mrjob/dataproc.py
```diff
@@ -0,0 +1,873 @@
...
+        ).execute()
+    
+    def _cancel_job(self, job_id):
+        return self.dataproc_client.jobs().cancel(
+            projectId=self._gcp_project,
+            region=self._gcp_region,
+            jobId=job_id
+        ).execute()
+
+    def _delete_job(self, job_id):
```
davidmarin: Oh, you can just delete a job. That's convenient! Might as well implement `_cleanup_job()` then.

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+        bucket_name, base_name = parse_gcs_uri(base_uri)
+
+        # allow subdirectories of the path/glob
+        if path_glob and not path_glob.endswith('/'):
+            dir_glob = path_glob + '/*'
+        else:
+            dir_glob = path_glob + '*'
+
+
+        fields_to_return = "nextPageToken,items(id,size,bucket,name,md5Hash,etag)"
```
davidmarin: Is there any additional expense to requesting this info, if we don't use it? I'm guessing the API call was designed for us to only ask for what we need.

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+        use the same read_file() method (in util.py) that the S3 and local versions are using.
+        But I can't figure out how to build the right file object to pass it to read_file().
+        Help appreciated.
+        """
+
+    def mkdir(self, dest):
+        """Make a directory. This does nothing on GCS because there are
+        no directories.
+        """
+        raise NotImplementedError
```
davidmarin: That's fine, but `pass`, don't raise an exception.

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+    def exists(self, path_glob):
+        """Does the given path exist?
+
+        If dest is a directory (ends with a "/"), we check if there are
+        any files starting with that path.
+        """
+        # just fall back on ls(); it's smart
+        try:
+            paths = self.ls(path_glob)
+        except:
```
davidmarin: Don't just swallow any exception here (basically, if the client raises an exception when you try to list a nonexisting path, catch that).

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+        """ TODO: too much common code between touchz and updload. Refactor.
+        Make an empty file in the given location. Raises an error if
+        a file already exists in that location."""
+        if self.exists(dest):
+            raise Exception("File already exists: " + dest)
+        media = http.MediaIoBaseUpload(io.BytesIO(""), "text/plain")
+        req = self.service.objects().insert( bucket=parse_gcs_uri(dest)[0], name=parse_gcs_uri(dest)[1], media_body=media)
+        req.execute()
+
+    def upload(self, local_path, destination_uri):
```
davidmarin: Yeah, this really should be part of the `Filesystem` API. But it should be named `put` (like the `hadoop fs` command).

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+    components = urlparse(uri)
+    if (components.scheme not in ("gs") or
+        '/' not in components.path):  # noqa
+
+        raise ValueError('Invalid GCS URI: %s' % uri)
+
+    return components.netloc, components.path[1:]
+
+
+if __name__ == "__main__":
```
davidmarin: This shouldn't be in the final version. :)

mrjob/fs/gcs.py
```diff
@@ -0,0 +1,224 @@
...
+class GCSFilesystem(Filesystem):
+    """Filesystem for Google Cloud Storage (GCS) URIs. Typically you will get one of these via
+    ``DataprocJobRunner().fs``, composed with
+    :py:class:`~mrjob.fs.ssh.SSHFilesystem` and
+    :py:class:`~mrjob.fs.local.LocalFilesystem`.
+    """
+
+    def __init__(self):
+        credentials = GoogleCredentials.get_application_default()
+        self.service = discovery.build('storage', 'v1', credentials=credentials)
```
davidmarin: Probably should call this `self._service` unless you're confident you want to expose it with this name.

mrjob/options.py
```diff
@@ -285,6 +282,74 @@ def add_hadoop_opts(opt_group):
...
+            '--gcp-zone', dest='gcp_zone', default=None,
+            help=('Zone to run Dataproc jobs in.  ')),
+
+        # TODO - mtai @ davidmarin - collapse cluster_id
+        opt_group.add_option(
+            '--gcp-cluster', dest='gcp_cluster',
+            help=('Dataproc cluster name - REQUIRED')),
+
+        # instance types
+        # TODO - mtai @ davidmarin - collapse with ec2_instance_type
```
davidmarin: Great, glad you're already thinking about this.

I think the Dataproc runner should just use general names for things (e.g. `num_instances`, `image_version`) from the get-go, and we should move the EMR runner over to these names.

davidmarin:
You'll need unit tests, of course (though it's better to get something working first). Let's talk about a testing strategy in person.

davidmarin:
Glad to see you're still working on this on a weekly basis! Planning to submit a pull request to fix your merge conflicts once I get mrjob v0.5.0 out.

mtai:
I think I'm ready for a code review :)  Thanks for all the py3k fixes.  I've been developing in Python 2.7 for a few years now.

Adding Dataproc support to mrjob
- Auto-generating bucket / cluster names
- Idle cluster termination (when --max-hours-idle passed)
- Added Dataproc docs
- Added Dataproc tests

davidmarin:
Merged this directly (along with some changes to comments and documentation). Thanks!

PR #1272, opened 2016-03-19
Replace use of deprecated 'hadoop dfsadmin' command

### LINKED ISSUE #1252: hadoop dfsadmin is deprecated
We use this to get the addresses of slave nodes, so we can fetch logs from them:

```
$ hadoop dfsadmin -report | grep ^Name | cut -f2 -d: | cut -f2 -d' '
DEPRECATED: Use of this script to execute hdfs command is deprecated.
Instead use the hdfs command for it.

172.31.16.204
```

Apparently, we should use the `hdfs` command instead (this is on the 4.0.0 AMI, though assuming this is for YARN in general).

ameya-pandilwar:
Fixes #1252

### DIFF
--- a/docs/guides/emr-advanced.rst
+++ b/docs/guides/emr-advanced.rst
@@ -230,7 +230,7 @@ If you have multiple nodes, you may have to :command:`scp` your identity file
 to the master node and use it to SSH to the slave nodes, where the core dumps
 are located::
 
-    hadoop@ip-10-160-75-214:~$ hadoop dfsadmin -report | grep ^Name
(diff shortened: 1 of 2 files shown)


davidmarin:
Oh yeah, there isn't a mock `hdfs` command.

Going to try to get this working for 0.5.1.

davidmarin:
Decided to get the slave IP addresses from the `ListInstances` API call instead, which means we won't need either `hadoop dfsadmin` or `yarn hdfs`. See #1345.

PR #1274, opened 2016-03-22
Fix typos in release notes

irskep:
Other places you might want to revisit:

* http://mrjob.readthedocs.org/en/latest/guides/runners.html#running-on-your-own-hadoop-cluster (describes Hadoop setup)
* http://mrjob.readthedocs.org/en/latest/guides/writing-mrjobs.html#multi-step-jobs (refers to `mr()` several times)


davidmarin:
Thanks!

PR #1295, opened 2016-04-07
Add --ssh-socks-proxy

jimjh:
Tests to come after I get some initial feedback!

This adds a `--ssh-socks-proxy` option to set up a SOCKS proxy using `ssh -D`. Note that this does not replace the existing SSH tunnel functionality, since having a `tunnel_url` to query is extremely useful.

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -802,3 +802,12 @@ SSH access and tunneling
     if True, any host can connect to the job tracker through the SSH tunnel
     you open.  Mostly useful if your browser is running on a different machine
     from your job runner.
+
+.. mrjob-opt::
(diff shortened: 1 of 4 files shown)


mtourne:
This is awesome! My organization has been needing this for a while.

jimjh:
Why we need this: the current tunnel does not allow the browser to view logs on worker nodes, whereas a SOCKS tunnel provides that _and_ allows us to use other applications such as `ganglia`.

davidmarin:
Can you try something for me? What happens if you run your job with `--ssh-bin 'ssh -D 2001'`? (Replace 2001 with your desired SOCKS proxy port.)

jimjh:
That worked perfectly.

PR #1327, opened 2016-06-08
Check current directory for config

jacobtomlinson:
Different projects may require different configuration. I find dropping a config file in my git repository is easier than setting an environment variable.

This PR adds the ability to create a file in the current directory called `mrjob.conf` which will be taken before `~/.mrjob.conf` but after `MRJOB_CONF`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -127,6 +127,7 @@ other advanced mrjob features, you'll need to set up ``mrjob.conf``. mrjob looks
 for its conf file in:
 
 * The contents of ``$MRJOB_CONF``
+* ``./mrjob.conf``
 * ``~/.mrjob.conf``
 * ``/etc/mrjob.conf``
 
--- a/docs/guides/configs-basics.rst
+++ b/docs/guides/configs-basics.rst
@@ -6,6 +6,7 @@ Config file format and location
 We look for :file:`mrjob.conf` in these locations:
 
 - The location specified by :envvar:`MRJOB_CONF`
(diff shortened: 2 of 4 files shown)


davidmarin:
Cool, I'm glad you were able to make this change work for you. It's really not typical to load configs from the current directory on UNIX (this can lead to some surprising situations), so I'm going to pass on this change, but thank you for the feedback.

A workaround for other people who want this behavior with out a code patch is `export MRJOB_CONF=mrjob.conf`.

jacobtomlinson:
Thanks for the feedback. 

For the record I disagree, this is a very common practice in UNIX software development. See CI/CD tools like Travis or Circle, task runners like Make or Grunt or package managers like Pip or Gem. Project specific configuration files go in the project files. 

That being said the environment variable is an acceptable work around, I'll just have to get a task runner to set the environment before calling Python.

davidmarin:
Oh, that's definitely true, but mrjob doesn't really have a concept of projects. In mrjob, the "project" is a single script file, which may or may not have other supporting files.

I hadn't really thought about it before, but you could make an argument for mrjob-script-specific config files (e.g. a `mr_your_job.conf` in the same directory as `mr_your_job.py`). Would that be helpful? If you wanted to have several scripts in the same directory share a common config, you could do this with `include`s.

jacobtomlinson:
Yeah we were trying to find a way of dealing with the concept of "projects" with mrjob. 

A script specific config could definitely be a good way to go!

PR #1390, opened 2016-08-07
Don't hide (potential) input files

### LINKED ISSUE #1200: uploaded files whose names start with `.`
I've noticed that Hadoop 1.0.3 (on the EMR AMI 2.4.9) can't see "hidden" input files (that is, those whose names start with `.`) on HDFS. You get an error like:

```
15/12/23 22:02:19 ERROR security.UserGroupInformation: PriviledgedActionException as:hadoop cause:org.apache.hadoop.mapred.InvalidInputException: Input path does not exist: hdfs://172.31.26.114:9000/user/hadoop/tmp/mrjob/mr_wc.hadoop.20151223.215251.093964/files/.bashrc
```

and then you're like, but it *does* exist, I can `hadoop fs -ls` it. Waaaaa!

We can work around this by having `UploadDirManager` deliberately avoid giving uploaded files names that start with `.`.

Should find out how widespread this problem is first (does it happen on S3? YARN? For files other than input files?).

davidmarin:
This updates `UploadDirManager` so that it won't give uploaded files a path that starts with `_` or `.`, since these are ignored by Hadoop's input readers. (Fixes #1200) This doesn't distinguish between input files and those that are used for other purposes, but it seems not to matter (your job never sees the upload path anyway).

This doesn't affect `WorkingDirManager`, so if your setup script etc. expects a filename to start with `_` or `.`, it still will.

### DIFF
--- a/mrjob/setup.py
+++ b/mrjob/setup.py
@@ -212,12 +212,14 @@ def parse_legacy_hash_path(type, path, must_name=None):
     return {'path': path, 'name': name, 'type': type}
 
 
-def name_uniquely(path, names_taken=(), proposed_name=None):
+def name_uniquely(path, names_taken=(), proposed_name=None, unhide=False):
     """Come up with a unique name for *path*.
 
     :param names_taken: a dictionary or set of names not to use.
     :param proposed_name: name to use if it is not taken. If this is not set,
                           we propose a name based on the filename.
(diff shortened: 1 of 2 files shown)


davidmarin:
Was able to run `mr_word_freq_count` on my `.emacs` file, yay.

Issue #1370, opened 2016-07-29
Spark should match Python version


davidmarin:
Basically, whenever we run Spark jobs (on any runner), we should explicitly set PYSPARK_PYTHON to `runner._default_python_bin()`.

(This should involve setting the configuration variables `spark.executorEnv.PYSPARK_PYTHON` and `spark.yarn.appMasterEnv.PYSPARK_PYTHON` through `--conf`)

Probably a good idea to have clarity on Python versions on recent 4.x AMI versions first (see #1365).

davidmarin:
Should also warn about the lack of Python 3 support on 3.x AMIs (not because Python 3 isn't installed, but because the Spark version is too old to support it).

davidmarin:
This is a stumbling block even for Python 2.7, since mrjob won't be bootstrapped for the default python (Python 2.6).

davidmarin:
fixed by #1446

PR #1440, opened 2016-09-17
bootstrap Spark

### LINKED ISSUE #1367: EMR: bootstrap Spark
On the 4.x AMIs, this means adding Spark to `emr_applications` if there are any spark/spark script steps (should just look for steps whose type starts with `spark`, so we can handle `spark_jar` steps).

On the 3.x AMIs this means running the bootstrap action `file:///usr/share/aws/emr/install-spark/install-spark`. Should double-check that this is available on all AMIs that claim to support Spark.

Alternatively, we could set `NewSupportedProducts.member.1.Name` to `Spark` in `api_params`. (If there's already a parameter starting with `NewSupportedProducts.member.1.`, we could try `2`, `3`, etc.)

davidmarin:
This automatically installs Spark on EMR if it looks like your job has Spark steps. Fixes #1367 

You can currently only run a job if your Python version matches `python` on your EMR AMI (since we bootstrap mrjob for whatever version of python matches the local version); this will change with #1370.

### DIFF
--- a/docs/guides/emr-opts.rst
+++ b/docs/guides/emr-opts.rst
@@ -504,6 +504,21 @@ and install another Python binary.
       no longer installs Python 3 on AMI version 4.6.0+ by default
 
 .. mrjob-opt::
+   :config: bootstrap_spark
+   :switch: --bootstrap-spark, --no-bootstrap-spark
+   :type: boolean
+   :set: emr
+   :default: (automatic)
+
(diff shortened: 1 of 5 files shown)


davidmarin:
Hand-tested `mrjob.examples.mr_spark_wordcount` on default (3.11.0) and 4.7.2 AMIs.

Issue #1372, opened 2016-07-29
integrate jobconf with Spark


davidmarin:
Configurations set with `jobconf` should be passed through to Spark with `--conf`. (Spark options all start with `spark.`, so it doesn't look like we need to keep these separate from Hadoop configuration options.)

davidmarin:
Users should be able to override `--conf` if they want to, so the priority should be, from lowest to highest:

 - matching Python version (#1370)
 - cmdenv (#1371)
 - jobconf (this ticket)
 - `--conf` set through `step_args`

davidmarin:
fixed by #1446

Issue #1371, opened 2016-07-29
integrate cmdenv with Spark


davidmarin:
Environment variables set through `cmdenv` should be passed along to Spark via `spark.executorEnv.*` and `spark.yarn.appMasterEnv.*`.

davidmarin:
`appMasterEnv` is only needed when running in cluster mode (`--deploy-mode cluster`). When running in client mode (the default), the "app master" is your Python script.

Seems harmless to set `spark.yarn.appMasterEnv.*` all the time; hopefully it doesn't cause annoying warnings.

davidmarin:
fixed by #1446

PR #1445, opened 2016-09-23
Spark python bin, cmdenv, jobconf

davidmarin:
Ensure that python binary (see #1370), cmdenv (see #1371), and jobconf (see #1372) are all passed to `spark-submit` with `--conf` options.


davidmarin:
whoops, wrong head branch!

PR #1449, opened 2016-09-30
add debug logging for pooling feature

jimjh:
This allows users to identify and troubleshoot pooling issues and control resource usage on EMR.

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -2757,20 +2757,25 @@ class EMRJobRunner(MRJobRunner, LogInterpretationMixin):
         key_cluster_steps_list = []
 
         def add_if_match(cluster):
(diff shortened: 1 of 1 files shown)


davidmarin:
Thanks. Updated your change to use indentation rather than repeating cluster ID, added remaining reasons a cluster could be rejected, and added "OK" when a cluster is accepted.

Also printed the entire, sorted list of usable clusters, and made this a debug log message rather than info.

davidmarin:
Just wanted to let you know that this change has already helped Yelp diagnose a cluster pooling issue. Thanks!

jimjh:
yay!

davidmarin:
Thanks for creating this; it just helped us debug a tricky pooling issue at Yelp. :)

Issue #1469, opened 2016-10-28
integrate libjars with Spark


davidmarin:
Should integrate `libjars` with Spark's `--jars` option.

Pretty sure this should work with Python steps as well; should double-check.

davidmarin:
Oh, sure, because PySpark can use Hadoop input/output formats. See [hadoopFile()](http://spark.apache.org/docs/1.4.1/api/python/pyspark.html#pyspark.SparkContext.hadoopFile).

PR #1471, opened 2016-10-30
Spark JAR support

### LINKED ISSUE #1377: SparkJarStep
Spark also runs Java JARs, and we should support this. These are very similar to `SparkScriptStep` except you can also set `main_class`.

This is mostly a matter of having something to hand-test with. I guess we should be fine using the examples that come with Spark.

davidmarin:
This lets you run JARs with `spark-submit`, which works almost like running Spark scripts, except that you have to specify a main class, and you can't use ``py_files``. Fixes #1377 

This doesn't add libjar support (see #1469).


davidmarin:
Whoops, wrong head, trying again.

PR #1483, opened 2016-11-12
Spark libjars

### LINKED ISSUE #1469: integrate libjars with Spark

### LINKED ISSUE #1481: jar steps should support libjars, jobconf
libjars and jobconf are generic options to the Hadoop command, but JAR steps don't support them.

davidmarin:
This adds libjars support to Spark (fixes #1469) and also adds some missing libjars and jobconf support to JAR steps (fixes #1481).

Still need to hand-test this.


davidmarin:
Got an example of how to use Spark with nicknack working.

Somewhat alarmingly, when I used MultipleValueOutputFormat (passing it `Text, IntWriteable` rather than `Text, Text`) it resulted in an error being logged, but EMR didn't consider it a failure.

Issue #1513, opened 2017-01-04
Warn about S3 output dir in wrong region


davidmarin:
If you give EMR an S3 output path that's in another region, your job fails with this error:

```
Exception in thread "main" com.amazon.ws.emr.hadoop.fs.shaded.com.amazonaws.services.s3.model.AmazonS3Exception: Moved Permanently (Service: Amazon S3; Status Code: 301; Error Code: 301 Moved Permanently; Request ID: 56451A04FCF24CCE), S3 Extended Request ID: uXkfEaU0FnL/FNZ+bvTulXyGXd0Sez8v6UA8vhtPkQmuC3XG/ivTE09cPMwJ6NgTU0VRU15+0Qs=
```

at least on AMI 4.8.2.

We should warn the user if they specify an output path in the wrong region.

davidmarin:
Having poked at this a bit, it seems to be an issue that exists in the 4.x AMIs but not the 5.x ones (so it won't happen by default). I think this is not work the extra code and tests.

davidmarin:
Code for this is in https://github.com/davidmarin/mrjob/tree/wrong-region-warning. As it is, it's not a very useful "feature" because the warning may or may not be relevant. Doing this right would take more researching the behavior of various EMR AMIs, and probably having it quit with an error, like we do if the output dir already exists.

davidmarin:
Also would need to fix tests: we're getting a `NoSuchBucket` error because the code now checks a bucket that it previously didn't, or checks it too soon in the process.

PR #1514, opened 2017-01-04
add --step-output-dir option

### LINKED ISSUE #263: allow saving inter-step output to s3 for emr jobs
in the current dev branch, _s3_step_output_uri and _s3_step_input_uri both make use of hdft:// paths for intermediate data between steps.

this is fine if a cluster is running with only master and core type instances. but if we wanted to add task instances (which can be dynamically added/removed from a running cluster) we no longer have access to hdfs on these nodes and cannot store to/load from hdfs on them. 

potentially we could decide this if there are any task instances running when a step is created, but it may be better to allow this as a command line option since task instances can be added at any time and it would be likely that someone would know ahead of time if they were planning on resizing their cluster in the future.

davidmarin:
This allows you to specify where to put the intermediate output of multi-step jobs (e.g. on S3, for future safekeeping). We don't want to do this by default (HDFS is faster), but this can be useful for debugging.

Fixes #263.

### DIFF
--- a/docs/guides/configs-all-runners.rst
+++ b/docs/guides/configs-all-runners.rst
@@ -233,6 +233,23 @@ Temp files and cleanup
     Don't stream output to STDOUT after job completion.  This is often used in
     conjunction with ``--output-dir`` to store output only in HDFS or S3.
 
+.. mrjob-opt::
+   :config: step_output_dir
+   :switch: --step-output-dir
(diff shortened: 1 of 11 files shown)


davidmarin:
Thought this was failing in hand-testing, but it was actually due to a cross-region bucket issue (see #1513).

Issue #1524, opened 2017-01-12
local mode should ignore failures from pre-filters


davidmarin:
Otherwise if you filter something through grep, it'll return exit status 1 if there are no matching lines in the input.

Whereas in real Hadoop, only the exit status of the last command in the pipeline matters.

davidmarin:
If `sh_bin` exists, it might make sense for local mode to sh-bin-wrap the command, like we do in the other runners. We can reserve local mode's simulated piping for Windows, etc.

davidmarin:
The local runner's architecture makes this surprisingly difficult!

davidmarin:
Looks like we already use sh-wrapping to bootstrap mrjob, so we might as well sh-bin-wrap the command piping the pre-filter into the mapper/reducer/etc.

Issue #1521, opened 2017-01-11
steps()' handling of *_pre_filter() is broken


davidmarin:
If `mapper_pre_filter()`, `combiner_pre_filter()`, etc. are defined in your job, `MRJob.steps()` should  call them to find out what the command is. Instead, it just includes the method in the step description.

Issue #1126, opened 2015-09-12
mapper_pre_filter is triggering java.lang.RuntimeException


NicolasEhrhardt:
I am running a job on a file containing very large json text. I need to filter out some lines since not all of them are relevant and I don't want the mapper to spend too much time parsing it when it's not a relevant entry.

Rather than discarding them in the mapper, it seemed like a good job for [`mapper_pre_filter`](http://mrjob.readthedocs.org/en/latest/guides/writing-mrjobs.html#cmd-filters): I am using something similar to `grep pattern`. Unfortunately, when using it, all mappers are throwing `RuntimeException`.

> java.lang.RuntimeException: PipeMapRed.waitOutputThreads(): subprocess failed with code 2
>     at org.apache.hadoop.streaming.PipeMapRed.waitOutputThreads(PipeMapRed.java:372)
>     at org.apache.hadoop.streaming.PipeMapRed.mapRedFinished(PipeMapRed.java:586)
>     at org.apache.hadoop.streaming.PipeMapper.close(PipeMapper.java:136)
>     at org.apache.hadoop.mapred.MapRunner.run(MapRunner.java:63)
>     at org.apache.hadoop.streaming.PipeMapRunner.run(PipeMapRunner.java:36)
>     at org.apache.hadoop.mapred.MapTask.runOldMapper(MapTask.java:441)
>     at org.apache.hadoop.mapred.MapTask.run(MapTask.java:377)
>     at org.apache.hadoop.mapred.Child$4.run(Child.java:259)
>     at java.security.AccessController.doPrivileged(Native Method)
>     at javax.security.auth.Subject.doAs(Subject.java:415)
>     at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1140)
>     at org.apache.hadoop.mapred.Child.main(Child.java:253)

I am assuming that's an issue with how piping is done for filtering entries before the mapper. (The job is running well without that prefilter).

davidmarin:
Probably is! What version of Hadoop are you using?

davidmarin:
Shoot, without more details (e.g. what your `grep` command line was), I can't duplicate this bug. Closing for now.

aeeilllmrx:
I am getting this issue as well. Would like to filter out certain lines from json file before the mapper step.

davidmarin:
Okay, I suspect this has something to do with how YARN handles pipes (or if they're allowed at all).

[...]

Issue #1061, opened 2015-06-26
*_pre_filter doesn't work on compressed files in local mode


davidmarin:
This is related to #567. Currently, in local mode, input paths are passed directly to the job, and thus its the job that handles decompression of .gz files, etc. This means that pre-filter commands will see the compressed data, when they're supposed to see uncompressed data (like in real Hadoop).

davidmarin:
Local mode currently falls back on `cat` to handle pre-filters. One problem with this is that there's no `cat` on windows (it's called `type`), which is why we only use it for pre-filters.

We should just make a mrjob equivalent (`python -m mrjob.cat`) that can handle compressed files the way Hadoop does, and use it whether there's a pre-filter or not.

PR #1525, opened 2017-01-21
A ton of local mode/pre-filter fixes

davidmarin:
Made several important fixes to local mode and pre-filters:

#1061: Local mode can now handle compressed input files passed through pre-filters. Every file that local mode pipes into another command first goes through `mrjob/cat.py`, which auto-decompresses compressed files based on their file extension. `mrjob/cat.py` depends on the standard library only, so we don't need to hack `PYTHONPATH` to ensure access to `mrjob`

#1126: This was a symptom of general brokenness with pre-filters, and should be fixed.

#1521: You can now override `*_pre_filter()` methods to get pre-filters in a single-step job. This used be totally broken.

#1524: Local mode used to report a failure if the pre-filter command reported failure. This can be a problem; for example BSD grep reports exit status 1 on no matching input, and there's no way to change this. Fixed this by doing the same thing we do in "real" Hadoop; using `sh -c` to pipe the pre-filter command into the job command. This made it possible to get rid of a lot of pretend-piping code.

Also, local mode no longer expects mrjobs to be able to take to take path arguments, so we're halfway to fixing #567.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,6 @@
 v0.5.8, 2017-01-?? -- ???
+ * moved mrjob.util.bunzip2_stream() to mrjob.cat
+ * moved mrjob.util.gunzip_stream() to mrjob.cat
  * deprecated mrjob.util.bash_wrap()
  * deprecated mrjob.util.tar_and_gz()
  * deprecated SSHFilesystem.ssh_slave_hosts()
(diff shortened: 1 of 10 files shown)


davidmarin:
Also added an example that uses pre-filters (`mrjob.mr_u_word_freq_count`)

davidmarin:
Travis CI seems to be having a bad day too. Merging.

PR #1541, opened 2017-02-16
use relative path for cat.py in local mode

davidmarin:
Local mode now uploads mrjob/cat.py (which handles uncompressing input files, to pipe them into the task) into the job's working directory and invokes them by relative path. This allows you to use a specially designed Docker script as your `python_bin`, since now all the files you need to get into Docker will be in the job's working dir.


davidmarin:
Whoops, this needs to be merged into the 0.5.x branch. Re-submitting.

PR #1558, opened 2017-03-16
SSH tunnel fallback

### LINKED ISSUE #1547: EMR: fetch job progress via SSH if SSH tunnel fails
It's possible, due to firewall weirdness, to be able to SSH into the master node but not to set up an SSH tunnel. If this happens, mrjob should fetch pages from the job tracker/resource manager using `ssh curl ...` rather than through the tunnel.

davidmarin:
mrjob can now get job progress whether or not you have a working SSH tunnel, as long as you can SSH into the master node. Fixes #1547


davidmarin:
Whoops, meant to merge this against `v0.5.x`. Retrying.

Issue #1304, opened 2016-04-24
use boto3, not boto


davidmarin:
`boto` is nearing end-of-life. Using `boto3` would allow us to access the EMR API directly, rather than having to come up with workarounds for every new API feature. It also returns ordinary data structures rather than objects.

davidmarin:
Ugh, to support this correctly, we're going to need to be able to unpack `boto`/XML API arguments passed through `--emr-api-params` and add them to the data structure we pass to `boto3`'s `run_job_flow()`.

At least we already have a decent handle on how this works, and have code that does something similar in `tests.mockboto` (`_api_params_to_emr_object`).

I don't think we actually need a new `emr_api_params` option for `boto3`; we just need to translate keys in `emr_api_params` that have `.` in them into the appropriate nested portion of the data structure.

davidmarin:
Should re-visit how we want to handle throttling too, e.g. #1391. Would be cool if we could get away without `RetryWrapper`. boto3's handling of throttling is probably better than boto's, but we still want to be bulletproof against socket errors etc.

davidmarin:
I *think* we can get away without a retry wrapper. Trying to figure out how to bump the max. number of retries that botocore uses.

davidmarin:
It looks like boto's retry configuration is buried several layers deep, and is very difficult to change programmatically.

It looks like we just need to catch `botocore.exceptions.ClientError` and then look at its `response` field. If `response.get('Error', {}).get('Code') == 'ThrottlingError'`, we want to retry.

davidmarin:
boto3 is pretty aggressive about finding configuration files; see http://boto3.readthedocs.io/en/latest/guide/configuration.html?highlight=configuration

One thing I noticed is that it'll use the region in your configs for IAM, which is unhelpful because IAM only works for `us-east-1`.

davidmarin:
Looks like `boto3` can figure out endpoints for unknown regions:

```python
>>> s3 = boto3.client('s3', region_name='lolcatnia')
>>> s3._endpoint
s3(https://s3.lolcatnia.amazonaws.com)
```

davidmarin:
Basically don't ever want to set `endpoint_url` unless the user specified it explicitly.

PR #1561, opened 2017-03-24
port S3 code from boto to boto3

davidmarin:
mrjob now uses `boto3` for all of its S3 needs. This fixes the S3 part of #1304 (only EMR left).


davidmarin:
Also snuck in a bit of code to make `to_lines()` pass through most file handles, apparently.

davidmarin:
Travis CI seems to be covered in molasses at the moment, so relying on my hand-running of the unit tests.

PR #1562, opened 2017-03-24
boto3 tzutc() fix

davidmarin:
Turns out if you want to interact with `boto3` `datetime`s, you need to specify a timezone. This updates `boto3` code (everything having to do with S3 and IAM) to specify `dateutil.tz.tzutc()` as the time zone.

`dateutil` is a dependency of `boto3`. Might want to add it explicitly.


davidmarin:
Again, relying on hand-run unit tests

Issue #1535, opened 2017-02-07
drop Python 2.6 support


davidmarin:
This would allow us to:
 * remove Python 2.6-specific workarounds (e.g. for `urlparse()`)
 * simplify Python version picking on EMR
 * use context managers for `BZ2File` and `TarFile`
 * use `timedelta.total_seconds()`
 * stop using `unittest2`
 * use `os.relpath()` rather than workarounds
 * don't need to nest `with` statements
 * use `shutil.make_archive()` rather than our own code

It would certainly be nice to use set literals and set and dict comprehensions, but there's not a compelling reason other than style to change this across the code base. Similarly, we can now un-nest `with` statements, but that's just a style issue.

davidmarin:
We could dump the workaround for #621, which is a 2.6-specific SSL issue.

davidmarin:
Dropped the workaround for #621 while porting EMR code to ``boto3`` (see #1304).

davidmarin:
Should also remove the Python 2.6 test target on Travis CI.

davidmarin:
According to http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-whatsnew.html, Python 2.6 isn't even installed by default on AMI 5.5.0 and later.

davidmarin:
This also says Python 2.7 "remains the system default" on 4.6.0. When did it start being the system default?

davidmarin:
The default on 4.0.0 is Python 2.6.9.

davidmarin:
4.2.0 is also on Python 2.6.9; 4.4.0 is on Python 2.7.10

davidmarin:
4.3.0 is on Python 2.7.10 too. That's the cutoff

PR #1576, opened 2017-04-19
finish porting to boto3

### LINKED ISSUE #1304: use boto3, not boto

davidmarin:
This ports all our EMR code and tests from `boto` 2 to `boto3`, and eliminates all references to `boto` code. Fixes #1304.

This also removes some utility code designed to do things that are now handled by `boto3` internally.

Removed `RetryGoRound` because the SSL issue it addresses only affects Python 2.6, which we're dropping support for (see #1535).


davidmarin:
This also eliminates `tests/suite.py` and includes `tests.` in the path of tests in `setup.py`'s test printout, make it easier to re-run individual failed tests.

PR #1620, opened 2017-07-11
Accept input paths with an s3a prefix

chetmancini:
More a cosmetic than functional change since `boto` is handling under the hood, but sometimes my jobs fail on the `ValueError` when I copy/paste `s3a` uris into input configuration. Since `s3n` is deprecated it makes sense to just support all the protocol prefixes.


davidmarin:
Looks good! Thanks.

PR #1636, opened 2017-08-24
Support for --exclude tags in report_long_jobs

bchess:
Add a feature to report_long_jobs to exclude clusters with specific tags.

Internal ticket MRJOB-169

### DIFF
--- a/mrjob/tools/emr/report_long_jobs.py
+++ b/mrjob/tools/emr/report_long_jobs.py
@@ -41,6 +41,8 @@ Options::
                         -us-west-1.amazonaws.com). You usually shouldn't set
                         this; by default mrjob will choose the correct
                         endpoint for each S3 bucket based on its location.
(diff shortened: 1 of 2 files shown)


mrjob/tools/emr/report_long_jobs.py
```diff
@@ -110,6 +133,8 @@ def _find_long_running_jobs(emr_client, cluster_summaries, min_time, now=None):
                      pending longer than this
     :param now: the current UTC time, as a :py:class:`datetime.datetime`.
                 Defaults to the current time.
+    :param exclude: A list of strings of the form TAG_KEY:TAG_VALUE
```
analogue: Looks like ```exclude``` got removed from the function signature.

bchess: ack, thx

PR #1661, opened 2017-10-03
Cast sleep_secs to numeric

velocirabbit:
`sleep_secs` is never cast from a String to a numeric value, so this throws a TypeError.


davidmarin:
Good catch! I'd prefer to fix the underlying issue (and add tests). See #1664.

velocirabbit:
Ah, didn't see that! Only saw the issue I ran into while trying to run jobs for a homework assignment. I'll just try to fix that underlying issue now, then.

PR #1709, opened 2017-11-20
Update parse.py to support s3a

laucia:
Update parse.py to the 0.6.0 version to allow using s3a while retaining the 0.5.X style configurations


davidmarin:
Sure. Not positive there will ever be a v0.5.12 release, but if there is, this will be in there.

PR #1718, opened 2017-12-27
Fix error

sariogonfer:
I received an exception when try to use a incorrect format config file.
When you use a bad formated config file, the messages used to instance
the TypeError exception raise another exception dues to actual runners
doesn't have a runner_alias attribute.


davidmarin:
Good catch. Thanks!

Issue #1338, opened 2016-06-22
handle local input dirs like Hadoop Streaming


davidmarin:
If you pass Hadoop Streaming a directory as input, it reads all non-"hidden" files (files whose names don't start with `_` or `.`) in that directory, but doesn't recurse into subdirectories (should double-check this).

It's confusing that we can't do the same thing with directories on the local filesystem. It'd be pretty easy to implement this.

(Thanks @rschroll for pointing this out!)

davidmarin:
The current workaround is to use a glob:

`python mr_your_job.py input_dir/*`

davidmarin:
The relevant issue about "hidden" input files is #1200.

Originally thought this behavior was part of Hadoop Streaming, but it's actually part of `FileInputFormat` (see http://stackoverflow.com/questions/19830264/which-files-ignored-as-input-by-mapper), which may or may not be used by other JARs.

rschroll:
In some additional testing, we found that local directories work only if they're missing their trailing slash.  That is,
`local/dir`: Works
`local/dir/`: Doesn't work
`local/dir/*`: Works

okomestudio:
I believe messages in this thread are mostly obsolete, as the implementation appears to have changed in v0.6.x. Nonetheless, I'll try using the thread instead of creating a new one.

I'm currently evaluating migration from v0.5.x to v0.6.1, and it looks like the behavior of input when supplied as a directory has changed in v0.6.x. In v0.5.x, I've been simply supplying a directory name (without slash as in `local/dir`) as input and things worked fine.

However, the generator method for input in v0.6.x has changed:

https://github.com/Yelp/mrjob/blob/master/mrjob/job.py#L663-L681

and, there, I think the method implementation has become inconsistent with the method docstring, and it now only accept stdin or files explicitly specified, as supplying a directory throws an IOError on https://github.com/Yelp/mrjob/blob/master/mrjob/job.py#L679.

Is this change intended? I'm not exactly sure why the support for directory input has been removed, and I believe the form `local/dir/*` only works because shell wildcard expansion, and I feel it is less elegant since it breaks symmetry with something like S3 input for which only prefix needs to be supplied.

davidmarin:
Shoot, this behavior is actually intentional and goes hand-in-hand with the new improved local/inline mode. The idea being that streaming tasks are really only supposed to read from stdin. It used to be that local mode would pass file paths to your job, but now it sets up stdin and stdout properly just like Hadoop Streaming.

[...]

PR #1723, opened 2018-01-24
Accept directories as input paths

okomestudio:
This fixes the issue I personally raised in #1338.


davidmarin:
Shoot, this isn't the right fix; the local/inline runners should handle directory paths, not make the job do it. The file you'll want to patch is `mrjob/sim.py`.

See https://github.com/Yelp/mrjob/issues/1338#issuecomment-362130882 for more details. Apologies for not responding before you wrote this patch.

PR #1729, opened 2018-01-30
Add option for cpu-core usage for local mode

qiuwei:
Add an option to specify the number of cpu cores while running in local mode.


davidmarin:
Neat, glad this works.

I was thinking of letting users set this through `--jobconf` (e.g. `--jobconf mapreduce.tasktracker.map.tasks.maximum=4`) but now that seems a bit obscure.

I *think* `num_cores` is the best name for this option (more clear than, say, `max_tasks` or `cpu_count`). Going to take a day to think about it.

davidmarin:
Thanks!

PR #1743, opened 2018-03-23
Dataproc Quickstart

### LINKED ISSUE #1589: improve Dataproc quickstart
The Dataproc quickstart in the docs is a bit outdated and a bit incomplete. Specificallly:

 * how do you add the Dataproc API? (the easiest way is the search box)
 * should you create named credentials or default ones? what about role?
 * you need to run `gcloud init`
   * and select a region
 * how do you feed the credentials JSON you downloaded into `gcloud`?

### LINKED ISSUE #1742: dataproc runner should work without gcloud util
It should be possible to run mrjob just by pointing `$GOOGLE_APPLICATION_CREDENTIALS` to a credentials JSON file; no need to install the `gcloud` util.

This is related to #1589 from M1.

davidmarin:
This brings the Dataproc Quickstart up to date (fixes #1589).

Also added a small tweak to the way we load credentials, so it's possible to use mrjob with Dataproc without installing `gcloud` (fixes #1742).

### DIFF
--- a/docs/guides/dataproc-quickstart.rst
+++ b/docs/guides/dataproc-quickstart.rst
@@ -3,30 +3,45 @@ Dataproc Quickstart
 
 .. _google-setup:
 
-Configuring Google Cloud Platform (GCP) credentials
----------------------------------------------------
+Getting started with Google Cloud
+---------------------------------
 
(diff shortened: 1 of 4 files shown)


davidmarin:
Looks like we need to instruct people to create a service account too.

PR #1752, opened 2018-04-16
mapper_raw() and input manifest

### LINKED ISSUE #754: binary input with manifest file
This would help with #753 and other attempts to support binary data.

Say you want to take the [US Census TIGER Shapefiles](http://www.census.gov/geo/maps-data/data/tiger-line.html) as the input to the first step of your job. You want each mapper to receive exactly one file, and you need to know each file's name so you can match up the `.shp` and `.dbf` files in your reducer. And you don't want to write any Java, or include any custom jars.

I _think_ mrjob could handle it this way:
- make a manifest file, with the URI of one input file per line.
- use `--jobconf mapred.max.split.size=1` to force Hadoop to pass one line to each mapper.
- in the wrapper script that we already use for setup commands:
  - read the URI from stdin
  - export jobconf environment variables (e.g. `map_input_file`) so that the job thinks it's reading directly from the file.
  - use `hadoop fs` to pipe the file into the wrapped command: `set -o pipefail; hadoop fs -cat <uri> | "$@"`
- The job reads directly from `self.stdin` in its `mapper_init()` step (see #753).

All `local` and `inline` modes have to do to emulate this is to not split input files.

davidmarin:
Allow jobs to parse arbitrary binary files by creating an input manifest. Fixes #754.


davidmarin:
Broke out writing documentation for this change into #1753.

PR #1761, opened 2018-04-30
chunked uploading on dataproc runner

### LINKED ISSUE #1404: Dataproc: control chunked uploading
Currently, the `dataproc` runner does a chunked upload, apparently using the default chunk size of `googleapiclient.http.MediaIoBaseUpload`, which is apparently [half a meg](https://github.com/google/google-api-python-client/blob/master/googleapiclient/http.py#L77).

According to the [GCS docs](https://cloud.google.com/storage/docs/best-practices#uploading), it's not recommended to use chunked uploads _at all_ if possible. Probably there's a maximum chunk size, but something closer to the 100MB we use on EMR would probably make more sense.

davidmarin:
Enabled the `cloud_upload_part_size` option on the dataproc runner, which is used to set `chunk_size` when uploading. Fixes #1404

### DIFF
--- a/mrjob/cloud.py
+++ b/mrjob/cloud.py
@@ -26,6 +26,7 @@ from subprocess import PIPE
 
 from mrjob.bin import MRJobBinRunner
 from mrjob.conf import combine_dicts
+from mrjob.py2 import integer_types
 from mrjob.py2 import xrange
 from mrjob.setup import WorkingDirManager
 from mrjob.setup import parse_setup_cmd
@@ -61,6 +62,7 @@ class HadoopInTheCloudJobRunner(MRJobBinRunner):
         'check_cluster_every',
         'cloud_fs_sync_secs',
         'cloud_tmp_dir',
+        'cloud_upload_part_size',
         'cluster_id',
         'core_instance_type',
         'extra_cluster_params',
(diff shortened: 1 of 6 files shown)


mrjob/cloud.py
```diff
@@ -61,6 +62,7 @@ class HadoopInTheCloudJobRunner(MRJobBinRunner):
         'check_cluster_every',
         'cloud_fs_sync_secs',
         'cloud_tmp_dir',
+        'cloud_upload_part_size',
```
mtai: cloud_upload_part_size_mb  ?

I didn't see any docs anywhere stating the units of this param

mrjob/cloud.py
```diff
@@ -119,14 +121,13 @@ def __init__(self, **kwargs):
...
 
-
-
     ### Options ###
 
     def _default_opts(self):
         return combine_dicts(
             super(HadoopInTheCloudJobRunner, self)._default_opts(),
             dict(
+                cloud_upload_part_size=100,  # 100 MB
```
mtai: cloud_upload_part_size_mb ?

davidmarin: Good point. I think `cloud_part_size_mb` would do it, though that's outside the scope of bringing Dataproc up to parity with EMR. Filed this as #1774.

PR #1792, opened 2018-05-30
Google idle timeout

### LINKED ISSUE #1705: Use Dataproc's built-in cluster termination
Dataproc has a feature to automatically terminate clusters that have been idle more than a certain number of minutes. We should use that rather than the buggy idle self-termination script (see #1702).

davidmarin:
Use Dataproc's built-in idle timeout feature rather than faking it with an initialization action. Fixes #1705.


mrjob/dataproc.py
```diff
@@ -1130,16 +1123,14 @@ def _cluster_create_kwargs(self):
...
-            gcs_init_script_uris.append(
-                self._upload_mgr.uri(_MAX_MINS_IDLE_BOOTSTRAP_ACTION_PATH))
-
         # NOTE - Cluster initialization_actions can only take scripts with no
         # script args, so the auto-term script receives 'mrjob-max-secs-idle'
         # via metadata instead of as an arg
         cluster_metadata = dict()
         cluster_metadata['mrjob-version'] = mrjob.__version__
+
+        # TODO: remove this once lifecycle_config is visible through
```
mtai: Remove?

lifecycle_config referenced on line 1196

davidmarin: This is specifically for the `mrjob-max-secs-idle` metadata.

We don't need this anymore for the initialization action script that handles idle timeout, but it's also currently the only way you can see idle timeout information for your job in the Google Cloud Console or through the `gcloud` util, so I thought it was worth leaving in for now. Once `lifecycle_config` becomes generally visible, we can get rid of the metadata variable.

davidmarin: Oh, the NOTE above this is stale though. Removed in my release branch. Thanks!

PR #1801, opened 2018-07-10
Respect termination protection to avoid boto ValidationError

brenns10:
If there is a cluster with Termination Protection enabled, and this batch attempts to terminate it, the batch fails with ValidationError and cannot continue to the next cluster.

This patch assumes that Termination Protection is enabled on purpose, and skips clusters which have it enabled, allowing the rest of the clusters in the account to be checked and terminated.

I also have a backport for v0.5.x, which would be super helpful to have released.

### DIFF
--- a/mrjob/tools/emr/terminate_idle_clusters.py
+++ b/mrjob/tools/emr/terminate_idle_clusters.py
@@ -200,12 +200,14 @@ def _maybe_terminate_clusters(dry_run=False,
             num_idle += 1
 
         log.debug(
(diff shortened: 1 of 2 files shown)


brenns10:
Here is the v0.5.x backport: https://github.com/Yelp/mrjob/compare/v0.5.x...brenns10:v0.5.x?expand=1

davidmarin:
This looks good. 

Going to push a workaround for Travis CI's Python 3.3 issues before merging this.

PR #1802, opened 2018-07-11
v0.5.x backport of Termination Protected bugfix

brenns10:
This backports #1801 for v0.5.x. It does it slightly differently, since I can't find a `TerminationProtected` attribute ever returned by boto. So rather than checking for termination protection ahead of time, we catch the ValidationError and check if it mentions termination protection - it's a bit hacky.


davidmarin:
(Tests pass on all the versions of Python we currently support in v0.5.x, so we're good.)

PR #1806, opened 2018-07-20
Drop old pythons

davidmarin:
These pythons are heavy and we should stop carrying them around.

This drops support for Python 2.6 and Python 3.3 from v0.5.x


davidmarin:
Also stop testing PyPy3 (which was never official supported)

Issue #1847, opened 2018-09-25
Permission Error [WinError 5] on windows


Drulludanni:
the reason for this is that the temp file that is created gets its write permissions modified by this piece of code in `mrjob/sim.py`

```
def _chmod_u_rx(path, recursive=False):
    """make *path* user readable and executable. If *recursive* is true,
    make *path* and everything inside it executable."""
    if recursive:
        for dir_name, _, file_names in os.walk(path, followlinks=True):
            for file_name in file_names:
                _chmod_u_rx(join(dir_name, file_name))
    else:
        if hasattr(os, 'chmod'):  # only available on Unix, Windows
            os.chmod(path, stat.S_IRUSR | stat.S_IXUSR)
```

Which will cause shutil.rmtree(path) to fail later when cleaning up the temp directory.

A simple example to see this failure is by running this piece of code
```import os
import  stat

filename = "test_file.txt"
with open(filename,"w") as file:
    file.write("aaaa")

os.chmod(filename, stat.S_IRUSR | stat.S_IXUSR)
os.remove(filename)
```
Which will also fail for the same reason on windows. 
is there a reason for chmodding the file? Is it possible to change the code such that
```
if hasattr(os, 'chmod') and os.name != 'nt':  # disabled for windows due to issues with deleting the temp folder
    os.chmod(path, stat.S_IRUSR | stat.S_IXUSR)```
so that it works on windows and doesn't crash + leave garbage behind in the temp folder?

davidmarin:
Oh, interesting. I'm guessing that on Windows, `rmtree()` expects the file to have `stat.S_IWUSR` to be able to remove it.

This code exists to imitate the way Hadoop sets permissions on distributed cache files. It won't break anything to disable it on Windows.

Would you like to submit a pull request? mrjob uses `platform.system() == Windows` to detect Windows, but otherwise your code looks good.

Drulludanni:
Yeah, I'll do it.

davidmarin:
Thanks! Sorry I'm not able to support Windows better.

PR #1849, opened 2018-09-25
Fixing Permission Error [WinError 5] on windows

Drulludanni:
#1847

### DIFF
--- a/mrjob/sim.py
+++ b/mrjob/sim.py
@@ -648,7 +648,8 @@ def _chmod_u_rx(path, recursive=False):
             for file_name in file_names:
                 _chmod_u_rx(join(dir_name, file_name))
     else:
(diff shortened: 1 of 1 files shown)


mrjob/sim.py
```diff
@@ -647,7 +648,7 @@ def _chmod_u_rx(path, recursive=False):
             for file_name in file_names:
                 _chmod_u_rx(join(dir_name, file_name))
     else:
-        if hasattr(os, 'chmod'):  # only available on Unix, Windows
+        if hasattr(os, 'chmod') and platform.system() != "Windows":  # only available on Unix
```
davidmarin: This comment isn't really honest. Something like, "this causes rmtree() to fail on Windows; see #1847" would be better (and it's perfectly fine to devote a whole line to the comment).

Issue #1376, opened 2016-07-30
Spark setup scripts on YARN


davidmarin:
Seems like `--setup` should work with (Python) Spark jobs. For example, if we need to call `make` on some Python code, that makes sense in Spark too.

`spark-submit` only runs shell scripts and Java JARs. In theory, we could make a dummy Python script, that wraps the setup wrapper script, that wraps the Python script. However, we should check what happens if we set `$PYSPARK_PYTHON` to, say, `sh -ex spark-setup-wrapper.sh python`; maybe it's fine as long as we ultimately run Python?

Also, it's not clear to me how Spark jobs are run. Possibly they all run in the same working directory? Need to figure out what level of locking we need.

Seems like the spark setup wrapper script should be similar but not identical to the one for streaming. For example, we don't need to emulate `--py-files`. We also _may_ want to emulate `--archive` if not running on `yarn` (see #1374), though that seems like an edge case.

davidmarin:
Should be totally fine and reasonable to bump this into a later milestone. I seriously doubt anyone currently using Spark is expecting to be able to wrap their Python script with arbitrary shell scripts.

davidmarin:
I'm pretty sure this isn't possible; you could run a shell script when you start your Spark job, but the pieces of it that run on other nodes are operating on serialized data and won't run the shell script there.

davidmarin:
This is totally possible; instead of wrapping the commands run by Hadoop streaming, you wrap the python binary. When Spark spawns sub-tasks, it's just invoking the Python binary (or whatever it thinks is the Python binary, which could be a wrapper script).

We'll still need "classic" setup scripts for non-Python tasks (e.g. `mapper_cmd()`).

davidmarin:
This is actually a pretty good selling point for `mrjob spark-submit`. Get setup scripts, get `--archives` even when you're not on YARN.

davidmarin:
We're going to need to use the master node setup script to get the wrapper script onto the cluster (unless we want to put the entire script into the command that we tell Spark is the Python binary).

davidmarin:
Also, we'll want probably to set `PYSPARK_DRIVER_PYTHON` because the wrapper script will be in a different location for the driver (absolute local path) than it will be for executors (current working directory, added with `--file`).

davidmarin:
[...]

PR #1893, opened 2018-12-04
Adds master_setup command

aaronabf:
For EMR runners, this new option provides functionality to run setup commands on the master node prior to running a step on EMR.

Example of use:

```
runners:
  emr:

    cmdenv:
      JOB_ENV: something

    bootstrap:
      - <some initial cluster configuration>

    master_setup:
      - sudo mkdir -p /tmp/$SOMETHING
      - sudo cp code.tar.gz /tmp/$JOB_ENV
      - <distributes code in cluster>

    upload_files:
      - code.tar.gz#code.tar.gz
```


davidmarin:
Thanks, but I'm hoping for a better use case before adding this feature. You can distribute code using `bootstrap` or `setup`. The reason I added the master node setup step was specifically to make JARs available to Hadoop (it didn't appear to be enough to put them on S3, at least on some AMIs).

aaronabf:
> You can distribute code using bootstrap or setup

However, what if the code we wish to run changes? This is what `setup` is intended for; however, as your documentation [states](https://pythonhosted.org/mrjob/guides/spark.html#no-setup-scripts): "you can’t wrap Spark jobs in setup scripts". As such our clusters using spark would require an entire new cluster for each code change.

davidmarin:
Ohhh, right.

I think the right remedy would be to enable setup scripts on Spark (see #1376). I'll prioritize that (and you're welcome to submit a pull request for that, if you like). I've already seen a hacky version of this (replacing `python_bin` with a custom script) in use at Yelp, so I know this approach works.

I'm really sorry to reject this pull request; your code looks good and has reasonably good test coverage. I'm just trying to keep from adding (and having to maintain) a potentially confusing and unnecessary feature when there's a simple, more intuitive way to accomplish the same thing.

PR #1895, opened 2018-12-04
Add new parameter emr_backoff

aaronabf:
This parameter is used instead of `check_cluster_every` during EMR throttling retries.

Previously, when throttled, any EMR api call would backoff for `check_cluster_every*1.5^{retry count}` seconds. Consider the default of 30; after 10 retries the next attempt would be in approximately 30 minutes. This backoff is too long for certain use-cases. As such we add functionality to adjust the throttling initial backoff. Therefore, if this is set to 5, after 10 retries the next attempt would be in only 5 minutes. In the current code we can only adjust `check_cluster_every`. This is insufficient as this also affects termination/step completion polling. Therefore we add `emr_backoff`.

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -311,6 +311,7 @@ class EMRJobRunner(HadoopInTheCloudJobRunner, LogInterpretationMixin):
         'ec2_key_pair_file',
         'emr_action_on_failure',
         'emr_api_params',
+        'emr_backoff',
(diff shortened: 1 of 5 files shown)


davidmarin:
Can you tell me more about your use case? I'd rather avoid adding more knobs to mrjob's backoff logic, but there might be another way if I can understand better what specific problem you're trying to solve.

PR #1923, opened 2018-12-27
Spark setup scripts on YARN

### LINKED ISSUE #1376: Spark setup scripts on YARN

davidmarin:
This makes `--setup` work when running Spark on YARN by wrapping the Python binary. Fixes #1376.

### DIFF
--- a/docs/guides/configs-all-runners.rst
+++ b/docs/guides/configs-all-runners.rst
@@ -384,7 +384,9 @@ Job execution context
 
     Name/path of alternate shell binary to use for :mrjob-opt:`setup` and
     :mrjob-opt:`bootstrap`. Needs to be backwards compatible with
-    Bourne Shell (e.g. ``'sh'``, ``'bash'``, ``'zsh'``).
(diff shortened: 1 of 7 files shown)


davidmarin:
Hand testing now.

davidmarin:
Okay, fixed issues that arose in hand-testing the Hadoop runner.

PR #1929, opened 2019-01-03
Bug Fix: Recursively merge extra_cluster_params dict

aaronabf:
The function `_cluster_kwargs` builds a dict of options to submit to AWS
for cluster creation. The last operation the function performs is a dict update
with any custom parameters the user adds. Since this update is not recursive,
if the user sets a nested option that mrjob specifies, then user will override
everything in that nested option. Therefore we must perform a recursive
update. For example, the following will only end up setting the master security
group and not the ec2 key pair nor subnet:

```
    ec2_key_pair: key_name
    subnet: subnet-foo

    extra_cluster_params:
      Instances.EmrManagedMasterSecurityGroup: sg-bar
```

### DIFF
--- a/mrjob/cloud.py
+++ b/mrjob/cloud.py
@@ -410,8 +410,16 @@ class HadoopInTheCloudJobRunner(MRJobBinRunner):
     def _add_extra_cluster_params(self, params):
         """Return a dict with the *extra_cluster_params* opt patched into
         *params*, and ``None`` values removed."""
+        def recursive_dict_merge(base, new_dict):
+           """Merged new_dict into base recursively at each value"""
+           for key in new_dict:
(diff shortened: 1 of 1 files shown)


davidmarin:
This is a good example. However, it doesn't match the code, which expects a nested dictionary (`Instances: EmrManagedMasterSecurityGroup: sg-bar`) rather than a dot-separated parameter name (`Instances.EmrManagedMasterSecurityGroup`).

I don't want recursive merging to be the default because the whole point of `extra_cluster_params` is to be a hammer that lets you do whatever you want, including blowing away parameters that mrjob sets. However, dot-separated parameter names (like you have in your example) would also address this use case.

Added a ticket for this (#1934). Will try to get it into the next release.

aaronabf:
That's reasonable. Thanks for taking a look :)

Issue #1945, opened 2019-01-09
Spark harness should support SORT_VALUES


davidmarin:
This is probably possible to implement with `sortBy()`.

PR #1964, opened 2019-01-24
spark_harness support SORT_VALUES

lydian:
#1945


lydian:
@davidmarin I can't put you on reviewers yet, so just ping you directly

davidmarin:
`MRJob`s expect `SORT_VALUES` to sort based on the encoded value (the line), not the decoded key-value pair, because that's what Hadoop Streaming would do. (It's also a bit more robust, since not all decoded values are sortable.)

lydian:
btw, I did test run the script with mr_sort_and_group.py example, and the result is correct.

Issue #1942, opened 2019-01-09
Spark harness should support passthrough args


davidmarin:
Basically the harness just needs a `--script-args` switch or similar.

PR #1968, opened 2019-01-25
spark_harness support passthrough arguments

lydian:
#1942 

Setting nargs=argparse.REMAINDER which seems to be the easiest way. 
However, the drawback is that the ordering of the passthrough argument would be important. 
'spark_harness_script.py class input output --compression-codec-class compression.class --remaining args --remaining --options'
will have correct compression-codec-class 
but 
'spark_harness_script.py class input output --remaining args --remaining --options --compression-codec-class compression.class'
will have --compression-codec-class compression.class in the passthru_args. 

I am not sure if this is acceptable. Let me know what you think. Thanks!

### DIFF
--- a/mrjob/spark/mr_spark_harness.py
+++ b/mrjob/spark/mr_spark_harness.py
@@ -33,14 +33,18 @@ class MRSparkHarness(MRJob):
             help=('Java class path of a codec to use to compress output.'))
 
         self.add_passthru_arg(
-            '--passthru-args',
+            '--job-args',
+            dest='job_args',
             default='',
             help=('The arguments pass to the MRJob. Please quote all passthru '
                   ' args so that they are in the same string')
         )
 
     def spark(self, input_path, output_path):
-        args = [self.options.job_class, input_path, output_path]
+        args = [
+            self.options.job_class,input_path, output_path,
+            '--job-args', self.options.job_args
+        ]
 
         if self.options.compression_codec:
             args.append('--compression-codec')
--- a/mrjob/spark/mr_spark_harness.py
+++ b/mrjob/spark/mr_spark_harness.py
@@ -32,6 +32,13 @@ class MRSparkHarness(MRJob):
             dest='compression_codec',
             help=('Java class path of a codec to use to compress output.'))
 
+        self.add_passthru_arg(
+            '--passthru-args',
+            default='',
+            help=('The arguments pass to the MRJob. Please quote all passthru '
+                  ' args so that they are in the same string')
+        )
+
     def spark(self, input_path, output_path):
         args = [self.options.job_class, input_path, output_path]
 
--- a/tests/spark/test_mrjob_spark_harness.py
+++ b/tests/spark/test_mrjob_spark_harness.py
@@ -56,8 +56,6 @@ class MRSortAndGroupReversedText(MRSortAndGroup):
 
 class MRPassThruArgTest(MRJob):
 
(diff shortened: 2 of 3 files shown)


davidmarin:
The spark harness script doesn't have to be user-friendly. Just having a switch named `--job-args` that takes a string would be fine. Take a look at `mrjob.util.shlex_split()` for how to parse out the arguments.

davidmarin:
Yep. In mrjob, the function to quote arguments is `mrjob.util.cmd_line()`.

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -159,6 +191,14 @@ def test_sort_values_sorts_encoded_values(self):
...
                      b=['bowling', 'balloon', 'baby']))
 
+    def test_passthru_args(self):
+        input_bytes = b'\n'.join([
+            b'to be or',
+            b'not to be',
+            b'that is the question'])
 
+        job = self._harness_job(
+            MRSortAndGroupReversedText, input_bytes=input_bytes,
```
davidmarin: Confused, I thought this would run MRPassThruArgs, which seems to be unused.

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -53,6 +54,30 @@ class MRSortAndGroupReversedText(MRSortAndGroup):
     INTERNAL_PROTOCOL = ReversedTextProtocol
 
 
+class MRPassThruArgTest(MRJob):
+
+    INTERNAL_PROTOCOL = ReversedTextProtocol
```
davidmarin: Unless `ReversedTextProtocol` is important to testing passthru args, you should just use the default protocols.

tests/spark/test_mrjob_spark_harness.py
lydian: oops, I forget to change it back while testing some other stuff.  updated!

tests/spark/test_mrjob_spark_harness.py
lydian: Nope, it is not important. I will just remove this

lydian:
@davidmarin would you mind doing this CR again? thanks!

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -53,6 +54,28 @@ class MRSortAndGroupReversedText(MRSortAndGroup):
...
+        self.add_passthru_arg('--chars', action='store_true')
+        self.add_passthru_arg('--ignore')
+
+    def mapper(self, _, value):
+        if self.options.ignore:
+            value = value.replace(self.options.ignore, '')
+        if self.options.chars:
+            for c in value:
+                yield c, 1
+        elif self.options.lines:
```
davidmarin: Where does `self.options.lines` come from?

mrjob/spark/mr_spark_harness.py
```diff
@@ -32,6 +32,13 @@ def configure_args(self):
             dest='compression_codec',
             help=('Java class path of a codec to use to compress output.'))
 
+        self.add_passthru_arg(
+            '--passthru-args',
```
davidmarin: Could you please rename this to `--job-args`, for consistency with `--job-class`? Having a passthru arg named `passthru_args` is just a bit too recursively confusing.

Also, please specify `dest` explicitly, for consistency with the rest of the `mrjob` library.

PR #1971, opened 2019-01-26
Spark harness tests

### LINKED ISSUE #1948: missing unit tests for Spark MRJob harness
We have a basic Spark harness (see #1941) but not a good way of unit-testing it. Probably what we need is a local Spark runner (see #1361).

davidmarin:
Added missing unit tests for Spark harness, including some recent features. Fixes #1948.

Found and fixed couple of issues with implementing `SORT_VALUES` (see #1945). Also changed `--compression-codec-class` to simply `--compression-codec` (see #1963).


davidmarin:
Should probably add back the test of running the harness *script* in Spark (rather than the harness script's `main()` function, which is as close as we can get in inline mode).

PR #1978, opened 2019-02-01
make mrjob_spark_harness handle combiners

stug:


mrjob/spark/mrjob_spark_harness.py
```diff
@@ -72,17 +72,43 @@ def _run_step(step, step_num, rdd, make_job):
...
         rdd = rdd.map(lambda k_v: m_write(*k_v))
 
+    if step_desc.get('combiner'):
+        combiner_job = make_job('--combiner', '--step-num=%d' % step_num)
+        c_read, c_write = combiner_job.pick_protocols(step_num, 'combiner')
+
+        # run the combiner
+        rdd = rdd.map(c_read)
+
+        def combiner_helper(pairs1, pairs2):
```
davidmarin: This would be a good place for a comment about why we expect a single item list to be the common case, and why we don't both trying to combine lists with multiple values.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -72,17 +72,43 @@ def _run_step(step, step_num, rdd, make_job):
...
+            if len(pairs1) == len(pairs2) == 1:
+                return list(
+                    combiner_job.combine_pairs(pairs1 + pairs2, step_num),
+                )
+            else:
+                pairs1.extend(pairs2)
+                return pairs1
+
+        rdd = rdd.map(lambda k_v: (k_v[0], k_v))
```
davidmarin: This would be a good place for a comment explaining that mrjob combiners might potentially behave differently for different keys, so we need to include the key in the "value"

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -72,17 +72,43 @@ def _run_step(step, step_num, rdd, make_job):
...
+        rdd = rdd.combineByKey(
+            createCombiner=lambda k_v: [k_v],
+            mergeValue=lambda k_v_list, k_v: combiner_helper(k_v_list, [k_v]),
+            mergeCombiners=combiner_helper,
+        )
+        rdd = rdd.flatMap(lambda k_to_k_v: k_to_k_v[1])
+
+        rdd = rdd.map(lambda k_v: c_write(*k_v))
```
davidmarin: If you did this write step before the `flatMap()`, you could produce the same `key_and_lines` format that the reducer expects. And then you could implement `SORT_VALUES` in the `flatMap()`, like the previous code did.

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -53,6 +54,20 @@ class MRSortAndGroupReversedText(MRSortAndGroup):
...
+        yield (word, sum(counts) - 1)
+        yield (word, 1)
+
+
+class MRWordFreqCountFailingCombiner(MRWordFreqCount):
+
+    unique_exception_str = '619dc867a1a34793b2157c258c859e78'
+
+    def combiner(self, word, counts):
+        raise Exception(self.unique_exception_str)
```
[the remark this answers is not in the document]
davidmarin: That's actually a pretty elegant hack. Might be a bit weird in the actual code, but perfectly reasonable in a test.

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -198,5 +215,24 @@ def test_passthru_args(self):
             input_bytes=input_bytes,
             extra_args=['--chars', '--ignore', 'to'])
 
+    def test_combiner_called(self):
```
davidmarin: This doesn't really belong in this test class, which is about comparing output between runs with and without the harness. I wouldn't worry about the ~5s overhead of creating another `SingleSparkContext` (and I could optimize it out anyhow).

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -198,5 +215,24 @@ def test_passthru_args(self):
...
+
+        with job.make_runner() as runner, \
+                self.assertRaises(Exception) as assert_raises_context:
+            runner.run()
+
+        exception_text = assert_raises_context.exception.__str__()
+        expected_str = MRWordFreqCountFailingCombiner.unique_exception_str
+        assert expected_str in exception_text
+
+    def test_bad_combiner(self):
```
davidmarin: Likewise, the contract with the inline runner is just that it imitates Hadoop, which may choose whether or not to run combiners. So comparing output isn't really fair. Instead, just include the expected output in the test case, and explain why you expect this output even though it doesn't match the word frequency count.

It'd probably be a good idea to note if there's another test case in this class that invokes combiners and tests them passively by comparing output.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -72,17 +72,43 @@ def _run_step(step, step_num, rdd, make_job):
...
+
+        # run the combiner
+        rdd = rdd.map(c_read)
+
+        def combiner_helper(pairs1, pairs2):
+            if len(pairs1) == len(pairs2) == 1:
+                return list(
+                    combiner_job.combine_pairs(pairs1 + pairs2, step_num),
+                )
+            else:
```
davidmarin: I don't see a test that tests what happens when combiners yield zero values, or two or more. Both of these would be good candidates for a test case that simply compares output with running the job the normal way.

However, you should *also* have a separate test case that ensures that if a combiner spits out two or more values, we don't run either of those values back through the combiner; we don't want to be accidentally running poorly designed combiners in O(n^2) time.

tests/spark/test_mrjob_spark_harness.py
davidmarin: I take that back; there's already another test that just uses `self._harness_job()`. This is fine where it is.

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -244,5 +267,54 @@ def test_passthru_args(self):
...
+        # since each word appears 3 times, resulting in 2 calls to
+        # combine_pairs.
+        with job.make_runner() as runner:
+            runner.run()
+
+            self.assertEqual(
+                dict(job.parse_output(runner.cat_output())),
+                dict(one=1003, two=1003, three=1003))
+
+    def test_combiner_yields_zero_values(self):
```
davidmarin: Really wanted to see a test of adding an empty list to a single-item list. A combiner that summed integer values (and declined to spit out a key-value pair if the value was zero) would test this, *and* you could compare its output to that of running the job without the harness on the inline runner.

davidmarin: This test is good too, wouldn't hurt to have both.

davidmarin:
Thank you!

PR #1989, opened 2019-02-12
Spark harness can run a range of steps

### LINKED ISSUE #1950: Spark harness should be able to run a range of steps
Currently the Spark MRJob harness only supports "pure" MRJobs. It should also be able to run jobs that have a combination of `MRStep`s and spark steps.

On the harness side, this mostly requires that we be able to run a range of steps.

davidmarin:
Added `--start-step` and `--end-step` arguments to the Spark harness, so it can deal gracefully with jobs with mixed step types. Fixes #1950.

Also simplified the code in `MRSparkHarness` that passes arguments straight through to the Spark harness, so we don't have to keep switches in sync in three different places.

### DIFF
--- a/mrjob/spark/mr_spark_harness.py
+++ b/mrjob/spark/mr_spark_harness.py
@@ -14,7 +14,13 @@
 """A wrapper for mrjob_spark_harness.py, so we can test the harness with
 the inline runner."""
 from mrjob.job import MRJob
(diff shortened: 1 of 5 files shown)


davidmarin:
Travis CI appears to be asleep, but tests pass locally on Python 2.7 and 3.7, so merging.

PR #1990, opened 2019-02-15
Reduce EMR API calls in cluster pooling

aaronabf:
This splits up `_usable_clusters` into two functions `_compare_cluster_setup` and `_check_cluster_state`. The original function would check that every aspect of the cluster are within the desired restraints. However this requires us to re-describe the setup of each cluster on each subsequent pooling attempt. This this setup doesn't change we are making superfluous EMR API calls, potentially causing more throttling. As such I split out the static setup checks into `_compare_cluster_setup` and the dynamic step checks into `_check_cluster_state` and only call `_check_cluster_state` when looking for a cluster in the second pass. This should greatly reduce the number of EMR API calls.

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -45,7 +45,6 @@ except ImportError:
 import mrjob
 import mrjob.step
 from mrjob.aws import _DEFAULT_AWS_REGION
-from mrjob.aws import EC2_INSTANCE_TYPE_TO_COMPUTE_UNITS
 from mrjob.aws import EC2_INSTANCE_TYPE_TO_MEMORY
 from mrjob.aws import _boto3_now
 from mrjob.aws import _boto3_paginate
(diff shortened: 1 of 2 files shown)


davidmarin:
Sorry for the delay, too me a while to read through and absorb this. Basically, it helps with the case where there aren't currently pooled clusters to join, but we're willing to wait and try again. (And it breaks down some fairly monolithic code into more understandable bits, which is always helpful.) Great!

(And if your change had broken anything, pretty sure the tests would have caught it.)

aaronabf:
> took me a while to read through and absorb this

Ya the changes aren't too drastic but the diff makes it quite hard to parse.

> Basically, it helps with the case where there aren't currently pooled clusters to join, but we're willing to wait and try again.

Exactly!

Thanks for taking the time to review and consider my change. I'm very glad to be contributing to this project :)

PR #1991, opened 2019-02-21
spark harness preserves partitioning

### LINKED ISSUE #1973: Spark harness doesn't preserve partitions
The Spark harness uses `flatMap()` to unpack the lines assigned to each key with `groupBy()`, but it needs to call `flatMap()` with `preservesPartitioning=True`.

This is an easy change to make, but we need to come up with a regression test that catches it.

davidmarin:
This ensures that the Spark harness doesn't inadvertently re-partition after it groups together key-value pairs for the reducer. Fixes #1973.

Also fixed and tested a loose end: the spark harness can handle combiners that run commands, but it does so by skipping them.

### DIFF
--- a/mrjob/spark/mrjob_spark_harness.py
+++ b/mrjob/spark/mrjob_spark_harness.py
@@ -249,7 +249,7 @@ def _shuffle_and_sort(rdd, sort_values=False):
     return rdd
 
 
-def _run_reducer(reducer_job, rdd, sort_values=False):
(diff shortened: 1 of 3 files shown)


davidmarin:
Also factored out the parts of the spark Harness than run the mapper, combiner, etc. renamed what is now `_discard_key_and_flatten_values()`, and added and updated a ton of comments.

PR #2000, opened 2019-02-27
empty *_bin opt always means to use the default

### LINKED ISSUE #1926: *_bin options can have zero arguments
We shouldn't allow binaries to be empty, because then we're trying to run the first argument to the binary as a command.

davidmarin:
Carefully updated the handling of every `*_bin` option so that if you set it to an empty value (e.g. `--sh-bin ''`) it uses the default.

This keeps us from having binaries with zero arguments (which can lead to weird results), and gives a convenient way to say, ignore the config file, just use the default.

Fixes #1926

### DIFF
--- a/mrjob/bin.py
+++ b/mrjob/bin.py
@@ -50,7 +50,6 @@ from mrjob.py2 import PY2
 from mrjob.py2 import string_types
 from mrjob.runner import MRJobRunner
 from mrjob.setup import parse_setup_cmd
-from mrjob.step import _is_spark_step_type
 from mrjob.util import cmd_line
 from mrjob.util import shlex_split
 from mrjob.util import unique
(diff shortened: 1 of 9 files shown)


davidmarin:
For reference, here's how things used to work:

 * `gcloud_bin, `sh_bin`, and `ssh_bin` were literally set to a default value rather than being defaulted by code at the time they were used
 * setting `gcloud_bin` to `[]` disabled use of SSH
 * setting `hadoop_bin` to `[]` complete disabled use of the `hadoop` binary (inadvertently using a feature that was designed to allow use of `ssh user@host hadoop ...` by the EMR runner)
 * setting `sh_bin` to `[]` caused an error on initialization
 * setting `ssh_bin` to `[]` (mostly) disabled use of SSH

davidmarin:
Didn't do anything with the `interpreter` opt as it was intended to allow running a binary by setting it to `''` (and we're getting rid of it anyway).

PR #2005, opened 2019-02-28
Bug fix: Add retrying to boto3 pagination.

aaronabf:
At some point mrjob introduced a bug where pagination is no longer
retried. This is due to the following:

```
pag = boto3_client.get_paginator(api_call) # <-- retries creation of paginator
pgen = pag.paginate(**api_params)          # <-- nothing needs to/does happen
for page in pgen: ...                      # <-- makes API calls but nothing
                                           #     wraps this
```

Unfortunately it is not easy enough catch exceptions afterwards, as any
exception to a python generator will end generation. Rather we find a slightly
odd fix, we wrap the pagination object itself (`pgen` in the above example).
Therefore all methods in the [PageIterator ](https://github.com/boto/botocore/blob/develop/botocore/paginate.py#L189) will be retried, which will handle
exceptions within the generator.

Note this also moves `_wrap_aws_client` and similar methods from `mrjob.fs.s3`
to `mrjob.aws`.


davidmarin:
Good catch!

I think you may not need to add special code to `RetryWrapper` to wrap `__iter__()`, and instead just can wrap the `PageIterator` object (which would wrap `self._make_request()`).

Also, accessing the `RetryWrapper`'s hidden backoff attribute is a little odd; why not just add a `_min_backoff` parameter to `_boto3_paginate()`, like we currently do with `_delay`?

aaronabf:
> and instead just can wrap the `PageIterator` object (which would wrap `self._make_request()`).

That was my original hope.

In the code, `paginator.paginate(**api_params)` returns an object of type `PageIterator`. This is what my change now wraps with the `RetryWrapper` (what you are also suggesting). However, without the `__iter__` in the wrapper the `__iter__` method itself will be wrapped, instead of the paginator's calls.

davidmarin:
Well, right, but what we need to wrap is API calls (e.g. `_make_request()`), and if we wrap them, any API errors that get bubbled up to the iterator are already non-recoverable. Or is there some other way that the API calls aren't being wrapped that I've missed?

aaronabf:
> if we wrap them, any API errors that get bubbled up to the iterator are already non-recoverable

Not sure if I follow. By wrapping them and retrying we prevent the iteration for ending. Could you explain further your concern?

davidmarin:
Sorry for the slow response.

Okay, so the problem is, calling `__iter__()` on the wrapped `PageIterator` returns a generator which, when we iterate over its, runs on the *unwrapped* object. What we really want is to be able to run `__iter__()` with a different `self`, specifically a retry-wrapped version of the original self.

I think there's a cleaner and more generalizable way to do this. Let me think about it.

davidmarin:
Okay, I think what we need to do is have `RetryWrapper` check if the method it wants to wrap satisfies `inspect.isgeneratorfunction()`, and if it does, rather than wrapping the method, it calls the method with itself (the `RetryWrapper`) as the `self` rather than the underlying object.

davidmarin:
(Calling a method with a different self is just `self.method.__func__(other_self, *args, **kwargs)`. In Python 2, it's `im_func` instead of `__func__`.)

davidmarin:
Sorry to put this off for so long. Will try to get (or something very similar) this into the next release.

aaronabf:
Oh sorry! I totally missed your followups. Thanks David!

davidmarin:
Looks like we can fix this by, instead of wrapping `get_paginator()` to retry transient errors, calling `get_paginator()` with the retry-wrapped client as `self`. From then on, the paginator will always be calling the retry-wrapped client.

This doesn't generalize as nicely as I hoped, but we can implement this by sub-classing `RetryWrapper` and overriding its `get_paginator()` method.

And yes, this makes my head hurt too.

PR #2006, opened 2019-02-28
Add cluster info file cache

aaronabf:
Read-through file cache of cluster info to reduce EMR API calls.

### DIFF
--- a/mrjob/cache.py
+++ b/mrjob/cache.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+# Copyright 2009-2019 Yelp and Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from contextlib import contextmanager
+from datetime import datetime
+import fcntl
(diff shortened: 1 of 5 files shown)


davidmarin:
It looks like you have a serious need to reduce your number of API calls or you wouldn't have built this. Caches are notoriously tricky to get right and I don't really want to be responsible for maintaining such a feature. Let's look at #1990 first?

aaronabf:
So the real reason why I introduced this is due to a new runner I've written, which I wasn't planning to put a revision up for until these precursors were addressed. This new runner bypasses the EMR Step API and schedules directly onto YARN. This allows us to run tons of smaller jobs on a single cluster. In just half an hour I have run over a hundred jobs on just 3 clusters using this runner. A major difficulty we have been facing, and in particular with this runner, is EMR API throttling. This cache was absolutely necessary for this new runner to operate.

davidmarin:
Oh, interesting! If you bypass the EMR API, how does pooling know whether a cluster is free; does the API still report that the cluster is busy even if you didn't launch the job through the API?

Have you tried launching commands via ssh (`ssh ... hadoop jar ...`)?

Also, the way EMR's Hadoop only runs one job at a time is specific to EMR. For example, Google Cloud Dataproc just gives you a Hadoop cluster, and you can run multiple jobs in parallel if you choose.

aaronabf:
Each job declares its expected resource restraints (memory, cores, etc.). Then during pooling we query each valid cluster's YARN resource manager API for various metrics and decide if there is enough room on the cluster for our job. 

> Have you tried launching commands via ssh (ssh ... hadoop jar ...)?

The runner actually does submit the job over ssh via spark-submit.

> Also, the way EMR's Hadoop only runs one job at a time is specific to EMR. 

This is why I call this the "YarnEMRJobRunner". Since our entire infrastructure is on AWS we found this to be a necessity :/

davidmarin:
Closing this for now.

PR #2009, opened 2019-03-04
spark_harness_process_increment_counter

lydian:
Still figuring out how to output the counter result so that it work with `parse_mr_job_stderr`. Send out CR as a early check for potential logic issue.

### DIFF
--- a/mrjob/spark/mrjob_spark_harness.py
+++ b/mrjob/spark/mrjob_spark_harness.py
@@ -14,9 +14,11 @@
 """A Spark script that can run a MRJob without Hadoop."""
 import sys
 from argparse import ArgumentParser
+from collections import defaultdict
 from importlib import import_module
 
 from mrjob.util import shlex_split
+from pyspark.accumulators import AccumulatorParam
 
 # tuples of (args, kwargs) for ArgumentParser.add_argument()
 #
@@ -50,9 +52,22 @@ _PASSTHRU_OPTIONS = [
 ]
 
 
+class CounterAccumulator(AccumulatorParam):
+
+    def zero(self, value):
+        return value
+
+    def addInPlace(self, value1, value2):
+        for key in value2:
+            value1[key] += value2[key]
+        return value1
+
+
 def main(cmd_line_args=None):
     if cmd_line_args is None:
         cmd_line_args = sys.argv[1:]
+    if stderr is None:
+        stderr = sys.stderr
 
     parser = _make_arg_parser()
     args = parser.parse_args(cmd_line_args)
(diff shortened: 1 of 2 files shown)


lydian:
Added unit test for increment_counter, note that I only print the counter at the very end. Looks like all of our use case is using counter for unit testing only. I am not going to figure out how to make `parse_mr_job_stderr` work for harness script.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -50,6 +52,25 @@
...
+
+    def zero(self, value):
+        return value
+
+    def addInPlace(self, value1, value2):
+        for key in value2:
+            value1[key] += value2[key]
+        return value1
+
+def print_counter_status(counter, output_stream):
```
davidmarin: Oh, I was going to say, there isn't really any particular reason to use the Hadoop streaming counter format and `parse_mr_job_stderr()`, and doing so might be a bit confusing. You can use whatever format is easy to understand and code. The other nice thing about this is that then mrjob can use the same code to parse Spark driver output that it currently uses to parse Hadoop driver output.

Personally, if I were to imitate a counter format, I'd use the one that the Hadoop binary uses to summarize counters after a job is run (which is more like what you're doing, printing out all the counters once a step completes).

Alternatively, treating the counters as output and printing them to a separate "counter output" directory (specified to the Spark harness with, say, `--counter-output-dir`), might be more robust. When you're running `spark-submit` directly, it's pretty trivial to get the Spark driver output, but when you're running inside a cluster (e.g. EMR), it takes a bit of fumbling.

lydian: I like the idea of having --counter-output-dir, I guess we can do both. Feels like counter info are more FYI in most use case, and people may not want to have a separate location for those info. Maybe we output the same format at the end and if people specify `--counter-output-dir` we also output the result to the folder. How do you think?

davidmarin: The spark harness isn't really meant to be used directly by the end-user anyway; they're supposed to use it through the Spark runner. So I think it would be fine to just output counters one way and assume the user doesn't care about the details.

lydian:
rebased to the latest master that has the .travis.yml fix

@davidmarin Please take a look when you have time, thanks!

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -13,9 +13,20 @@
...
+import json
 from io import BytesIO
 from os.path import join
+from collections import defaultdict
+from contextlib import contextmanager
+from io import StringIO
+from tempfile import gettempdir
+from shutil import rmtree
 
+from mrjob.examples.mr_word_freq_count import MRWordFreqCount
```
davidmarin: There probably shouldn't be two of this line (but I always clean stuff like this out at release time anyhow).

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -47,9 +50,31 @@
         dest='compression_codec',
         help=('Java class path of a codec to use to compress output.'),
     )),
+    (['--counter-output-dir'], dict(
+        default=None,
+        dest='counter_output_path',
```
davidmarin: `dest` should match the switch (it should be `counter_output_dir`)

(argparse would also do this automatically if you didn't set `dest`, but In `mrjob` I always set the `dest` explicitly because I like to be able to grep for things.)

lydian: ah I rename it at some point, and forget to change everything. Thanks for finding it

PR #2010, opened 2019-03-08
 YARM EMR Runner

aaronabf:
This new runner bypasses the EMR Step API and schedules directly via YARN. This allows us to run a plethora of smaller jobs on a single cluster. In just half an hour I have run over a hundred jobs on just 3 clusters using this runner.


davidmarin:
Closing this for now. I think adding the ability to work over SSH to the Hadoop runner would be a more general solution that would address your problem.

Issue #1953, opened 2019-01-12
Spark harness should control number of output files


davidmarin:
This can be done with `coalesce()`. The Spark runner might do this automatically in response to configuration properties.

davidmarin:
`coalesce()` has the drawback that it can't increase the number of files. What we're really trying to do is imitate the Hadoop config property that sets the number of reducers. Adding `numPartitions` to `combineByKey()`/`groupBy()` and preventing re-partitions from there until output is a more robust approach.

davidmarin:
Oops, wrong ticket number on #2028.

PR #2016, opened 2019-03-11
Limit number of  output files in spark_harness_script

lydian:
#1953

### DIFF
--- a/mrjob/spark/mrjob_spark_harness.py
+++ b/mrjob/spark/mrjob_spark_harness.py
@@ -89,7 +89,7 @@ def main(cmd_line_args=None):
     parser = _make_arg_parser()
     args = parser.parse_args(cmd_line_args)
 
-    if args.num_reducers and args.num_reducers <= 0:
+    if args.num_reducers is not None and args.num_reducers <= 0:
         raise ValueError(
             'You can only configure num_reducers to positive number.')
 
@@ -175,7 +175,7 @@ def main(cmd_line_args=None):
             )
 
 
-def _run_step(step, step_num, rdd, make_job, num_reducers):
+def _run_step(step, step_num, rdd, make_job, num_reducers=None):
     """Run the given step on the RDD and return the transformed RDD."""
     step_desc = step.description(step_num)
     _check_step(step_desc, step_num)
@@ -214,7 +214,7 @@ def _run_step(step, step_num, rdd, make_job, num_reducers):
             rdd, sort_values=sort_values, num_reducers=num_reducers)
 
     if reducer_job:
-        rdd = _run_reducer(reducer_job, rdd)
+        rdd = _run_reducer(reducer_job, rdd, num_reducers=num_reducers)
 
     return rdd
 
@@ -343,7 +343,7 @@ def _shuffle_and_sort(rdd, sort_values=False, num_reducers=None):
     return rdd
 
 
-def _run_reducer(reducer_job, rdd):
+def _run_reducer(reducer_job, rdd, num_reducers=None):
     """Run our job's combiner, and group lines with the same key together.
 
     :param reducer_job: an instance of our job, instantiated to be the mapper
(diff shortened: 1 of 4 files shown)


mrjob/spark/mrjob_spark_harness.py
```diff
@@ -140,7 +146,7 @@ def make_job(*args):
             )
 
 
-def _run_step(step, step_num, rdd, make_job):
+def _run_step(step, step_num, rdd, make_job, num_reducers):
```
davidmarin: It would be helpful for readability to say `num_reducers=None` so that people can see at a glance that it doesn't have to be a number.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -279,20 +292,23 @@ def combiner_helper(pairs1, pairs2):
...
+                        are sorted (by their encoded value)
+    :param num_reducers: limit the number of paratitions of output rdd, which
+                         is similar to mrjob's limit on number of reducers.
     :return: an RDD containing "reducer ready" lines representing encoded
              key-value pairs, that is, where all lines with the same key are
              adjacent and in the same partition
     """
-    rdd = rdd.groupBy(lambda line: line.split(b'\t')[0])
+    rdd = rdd.groupBy(
+        lambda line: line.split(b'\t')[0], numPartitions=num_reducers)
```
davidmarin: For this, and `combineByKey()`, once we set a specific number of partitions, we're pretty much committed to no re-partitions until the end of the step. I'd expect to see `preservesPartitioning=bool(num_reducers)` on every operation that follows this one.

Once you do this, you'd probably want to add a test case or two to `tests.spark.test_mrjob_spark_harness.PreservesPartitioningTestCase`.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -57,6 +57,12 @@
             'An empty directory to write counter output to. '
             'Can be a path or URI.')
     )),
+    (['--num-reducers'], dict(
+        default=None,
+        dest='num_reducers',
```
davidmarin: You should probably add a check in `main()` that ensures that `args.num_reducers` is `None` or a positive number. Setting num reducers to 0 is used as a way to tell Hadoop to skip the reduce step, and we should be explicit that the harness doesn't emulate that.

lydian: thanks! Didn't aware that!!! I'll update this.

mrjob/spark/mrjob_spark_harness.py
```diff
@@ -263,6 +280,7 @@ def combiner_helper(pairs1, pairs2):
         createCombiner=lambda k_v: [k_v],
         mergeValue=lambda k_v_list, k_v: combiner_helper(k_v_list, [k_v]),
         mergeCombiners=combiner_helper,
+        numPartitions=num_reducers
```
davidmarin: expecting to see `preservesPartitioning=bool(num_reducers)` in the call to `mapValues()` below and inside `_discard_key_and_flatten_values()`

tests/spark/test_mrjob_spark_harness.py
```diff
@@ -488,32 +545,40 @@ def _test_shuffle_and_sort(self, sort_values):
...
         reducer_job = Mock()
         reducer_job.pick_protocols.return_value = (Mock(), Mock())
 
-        final_rdd = _run_reducer(reducer_job, rdd)
+        final_rdd = _run_reducer(reducer_job, rdd, num_reducers=num_reducers)
         self.assertEqual(final_rdd, rdd)  # mock RDD's methods return it
 
         called_mapPartitions = False
-        for name, args, kwargs in rdd.method_calls:
+        expected_preservePartitioning_vals = [
```
davidmarin: I'd prefer if you didn't make the test expect exactly three calls; it should be possible to refactor `_run_reducer()` a bit without having to update this test. The test is just trying to check `groupBy()` and every call after.

mrjob/spark/mrjob_spark_harness.py
lydian: http://spark.apache.org/docs/latest/api/python/pyspark.html?highlight=mapvalues#pyspark.RDD.mapValues

mapValues retains the original RDD’s partitioning therefore we don't need to (and also unable to) specify preservesPartitioning

tests/spark/test_mrjob_spark_harness.py
lydian: I slightly modified the call, so that we don't limit to 3  calls.

davidmarin:
Looks good, just waiting for tests to pass.

davidmarin:
I'll take care of resolving the merge conflict.

PR #2021, opened 2019-03-15
Rename files in working dir on all Spark masters

### LINKED ISSUE #1922: --files won't work properly on non-YARN Spark
Our code assumes that `--files foo#foo` works the same as `--files foo`. This is true on Hadoop, but most Spark installations (at least, `local` mode and `Mesos`) don't allow hash paths at all.

### LINKED ISSUE #2014: use fs.mkdir() to create buckets
Currently trying to generalize code to upload local files to the cloud, and stumbling on bucket creation. Our fs should just create buckets as needed when we `fs.mkdir()` inside them. We can specify the location of newly created buckets when the fs is initialized.

davidmarin:
This makes `--files` support Hadoop distributed cache syntax (`path#name`) on non-YARN Spark masters (e.g. Mesos). Fixes #1922.


davidmarin:
This pull request makes me a bit nervous because it does so many things at once. Going to try to split out some of these things (e.g. `mkdir()` making buckets) into separate pull requests.

davidmarin:
This is by now a fairly simple change that I mostly trust, going to merge it so people can try it out.

PR #2023, opened 2019-03-19
integrate Spark harness counter simulation into Spark runner

### LINKED ISSUE #1955: Spark harness/runner should simulate counters
The Spark harness should patch `MRJob`s' `increment_counter()` method to update an [accumulator](https://spark.apache.org/docs/latest/rdd-programming-guide.html#accumulators) or similar, and have some way of reporting that back to the Spark runner.

davidmarin:
This implements `runner.counters()` and ensures that non-empty counters are logged.

When the Spark harness runs a multi-step job, it now creates a separate accumulator for each step of the job, rather than combining all the counters into a single nested dictionary.

Fixes #1955.


mrjob/spark/runner.py
```diff
@@ -160,7 +168,14 @@ def _pick_spark_tmp_dir(self):
...
     def _default_step_output_dir(self):
-        return posixpath.join(self._spark_tmp_dir, 'step-output')
+        return self.fs.join(self._spark_tmp_dir, 'step-output')
+
+    def _counter_output_dir(self, step_num):
+        return self.fs.join(
+            self._spark_tmp_dir, 'counter-output-step-%d' % step_num)
+
+    def counters(self):
+        return deepcopy(self._counters)
```
lydian: Kind of curious why this returning a copy of counters, I am guessing you want to avoid the number being changed accidentally by client?

davidmarin: Yeah. I mean, it can't be that important or I would have changed it ages ago, but I definitely don't mean for `self._counters` to be editable.

Issue #1977, opened 2019-02-01
add put() method to fs


davidmarin:
Every runner has some way of uploading local files to the filesystem, but `mrjob.fs.base.Filesystem` doesn't provide a consistent interface (it's basically read-only).

The Spark runner (see #1940) actually needs this because it supports all the filesystems. We just need to be able to upload a local file to a remote URI (no recursive behavior necessary).

`GCSFilesystem` already has a `put()` method, so we should follow that interface.

PR #2027, opened 2019-03-21
remove part_size kwarg from fs.put()

davidmarin:
`part_size` (for multi-part upload on S3 and GCS) is now set when filesystems are initialized. Also removed unnecessary shim methods for uploading, and moved the code for determining part size into `mrjob.runner`.

This is a mid-release change, so there's no issue number, but this is relevant to #1977 (adding `put()` to fs).


davidmarin:
(This pull request helps pare down #2021, which does too many things at once.)

PR #2028, opened 2019-03-21
fs.mkdir() creates buckets

### LINKED ISSUE #1953: Spark harness should control number of output files

davidmarin:
`fs.mkdir()` will now implicitly create buckets on the GCS and S3 filesystems. Region and (on GCS) object TTL are now set at fs init time. Fixes #1953.

This change allowed us to simplify a fair amount of runner code having to do with creating buckets and uploading files.


davidmarin:
This pull request helps pare down #2021, which is too big.

PR #2063, opened 2019-05-13
Match increment_counter arguments in spark harness with base job

### LINKED ISSUE #2060: unexpected kwarg in spark harness increment_counter()
The Spark harness's emulation of `increment_counter()` isn't perfect. It has args in the right order, but the names don't match:

```
TypeError: increment_counter() got an unexpected keyword argument 'counter'
```

qui:
Fixes #2060.

The updated test `tests.spark.test_harness.SparkHarnessOutputComparisonTestCase.test_increment_counter` failed before the change to the harness, and now passes. The non-spark tests using MRCountingJob still pass.


davidmarin:
Looks good, just waiting for the tests to pass.

PR #2107, opened 2019-10-02
make report_long_jobs report jobs in WAITING state

stug:


davidmarin:
This isn't what `mrjob report-long-jobs` is for; it's for catching long-*running* jobs. It's meant as a complement to `mrjob terminate-idle-clusters`, which catches clusters that have been in the `WAITING` state for a long time.

PR #2128, opened 2019-11-13
stop reading pre-v0.6.0 pool names

### LINKED ISSUE #1815: mrjob-audit-usage makes unnecessary call to `ListBootstrapActions`
Currently, `audit-emr-usage`calls `ListBootstrapActions` for each cluster.

The only thing we use this information for is to determine which pool the cluster is in, for clusters launched by "legacy" (pre-0.6.0) versions of mrjob. If the cluster is tagged with `__mrjob_version` (introduced in v0.6.0), we know that we can get the pool name (if any) from the cluster's tags, which is available from `ListCluster`, so there's no need to fetch the bootstrap actions.

davidmarin:
`mrjob audit-emr-usage` no longer reads pool names from clusters launched by pre-0.6.0 versions of mrjob. Fixes #1815.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -27,6 +27,7 @@ v0.7.0, 2019-10-?? -- fall cleaning
    * arguments to CompositeFilesystem constructor
    * MRJobLauncher class
    * the `mrjob run` command
+ * mrjob audit-emr-usage no longer reads pre-v0.6.0 cluster pool names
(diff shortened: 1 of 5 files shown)


davidmarin:
This saves an unnecessary API call to `ListBootstrapActions`, which was only used to read legacy pool names.

PR #2134, opened 2019-12-16
Set VisibleToAllUsers to true by default

### LINKED ISSUE #2131: clusters should always be visible to all users
We removed the `visible_to_all_users` option in mrjob v0.7.0, on the theory that it (a) wasn't very useful to hide clusters and (b) [according to the EMR API docs](https://docs.aws.amazon.com/emr/latest/APIReference/API_RunJobFlow.html), it defaults to `true` anyway, so we don't even have to bother sending it to the API anymore.

Whether because of `boto3` or the API itself, it turns out the that default is actually `false`, meaning all jobs launched by mrjob v0.7.0 are hidden from other users.

We should explicitly set `visible_to_all_users` to true when calling `run_job_flow()`. If users of mrjob desire to hide their clusters from other IAM users, they can always use `--extra-cluster-param VisibleToAllUsers=false`.

88manpreet:
Fixes #2131.

### DIFF
--- a/tests/test_emr.py
+++ b/tests/test_emr.py
@@ -465,6 +465,17 @@ class VisibleToAllUsersTestCase(MockBoto3TestCase):
         cluster = self.run_and_get_cluster()
         self.assertEqual(cluster['VisibleToAllUsers'], True)
 
+    def test_overridden_false(self):
+        cluster = self.run_and_get_cluster(
+                            '--extra-cluster-param', 'VisibleToAllUsers=false')
+        self.assertEqual(cluster['VisibleToAllUsers'], False)
+
+    def test_overridden_true(self):
+        cluster = self.run_and_get_cluster(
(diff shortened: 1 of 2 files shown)


tests/test_emr.py
```diff
@@ -459,6 +459,19 @@ def test_must_be_integer_in_config(self):
             self.assertRaises(ParamValidationError, self.run_and_get_cluster)
 
 
+class VisibleToAllUsersTestCase(MockBoto3TestCase):
+
+    def test_defaults(self):
+        cluster = self.run_and_get_cluster()
+        self.assertEqual(cluster['VisibleToAllUsers'], True)
+
```
davidmarin: You should also test `--extra-cluster-param VisibleToAllUsers=false`

88manpreet: Added.

PR #2135, opened 2019-12-18
add runner initialization to debug messages

### LINKED ISSUE #2129: add initialization of runner to debug messages
Currently writing a walkthrough of how mrjob works, and realized it would be helpful to have mrjob show which runner it's invoking, with with keyword arguments.

davidmarin:
Makes it a bit easier to understand how runners work. Fixes #2129


davidmarin:
Also makes `_default_opts()` a classmethod (though this isn't user-facing).

PR #2137, opened 2019-12-18
emulate --archives on non-YARN Spark

### LINKED ISSUE #1993: emulate archives on Spark using setup scripts
Spark supports uploading files to your job's working dir, but only Spark on YARN can support unpacking archives into directories in the job's working dir.

We should be able to emulate unpacking archives with a snippet of shell script, just like we provide archives in bootstrap scripts (see `mrjob/cloud.py`).

davidmarin:
If we're on a version of Spark that doesn't support ``--archives`` (basically, any non-YARN master), we upload the archives as files and then unpack them ourselves using a setup script. Fixes #1993.

This gives people another good reason to use `mrjob spark-submit`; it allows them to migrate from YARN to another Spark master without giving up support for archives or renaming files.


mrjob/runner.py
```diff
@@ -1345,21 +1365,31 @@ def _upload_args(self):
...
-            always_use_hash=True):
+            always_use_hash=True, emulate_archives=False):
         args = []
 
         file_hash_paths = list(
             self._file_arg_hash_paths(files,
                                       always_use_hash=always_use_hash))
+
+        # if emulating --archives, upload archives with files (we'll unpack
+        # them later with a setup script
```
88manpreet: Minor: Closing bracket in the end ')'

setup.py
```diff
@@ -38,32 +44,40 @@
...
             'ujson',
             'warcio',
         ],
         'zip_safe': False,  # so that we can bootstrap mrjob
     }
 
-    # Google libs don't install on Python 3.4. Which is fine, the only
-    # reason we support Python 3.4 at all is to support earlier
-    # AMIs on EMR. See #2090
-    if sys.version_info[0] == 2 or sys.version_info >= (3, 5):
```
88manpreet: Curious why this condition is removed?

mrjob/tools/spark_submit.py
```diff
@@ -26,6 +26,10 @@
    added ``local``, ``spark`` runners, made ``spark`` the default (was
    ``hadoop``)
 
+.. versionchanged:: 0.7.1
```
88manpreet: Should it be 0.7.2 or 0.7.2.dev0?

davidmarin: Should be v0.7.2.dev0. This is fixed in master

PR #2143, opened 2019-12-31
don't copy SSH key to master

### LINKED ISSUE #1209: don't copy SSH key
Currently, to SSH into EMR slave nodes, we first copy our SSH key to the master node.

It would be cleaner and more secure to use SSH agent forwarding. This might require a dummy config file, which we already make.

davidmarin:
When we needed to SSH into non-master nodes on EMR, we used to copy our key to the master node (which is potentially a security issue). This changes it so we add the key to the SSH agent instead. Fixes #1209.


88manpreet:
+1

tests/mock_boto3/case.py
```diff
@@ -16,6 +16,7 @@
 import itertools
 import os
 import shutil
+import stat
```
88manpreet: Missed this earlier. Where are we using this?

mrjob/fs/ssh.py
```diff
@@ -88,12 +90,17 @@ def _ssh_cmd_args(self, address, cmd_args):
                 key_pair_file = self._remote_key_pair_file
```
88manpreet: I think we are never using this, right? (when i !=0)

tests/mock_boto3/case.py
davidmarin: whoops, we're not. you can delete it (or remove it in pre-release pyflaking)

mrjob/fs/ssh.py
davidmarin: Oh, right, `self._remote_key_pair_file` is unused and should be removed.

PR #2152, opened 2020-03-24
fix setup.py and tests so that Python 3.4 and 3.8 work

davidmarin:
Python 3.4 needs to not install the latest version of PyYAML (3.4 has reached end-of-life; we're only keeping it around to support old AMIs)

`pyspark` doesn't currently support Python 3.8, so we need to not install it when testing, and we need to properly skip all the tests that depend on `pyspark`.


setup.py
```diff
@@ -38,32 +44,40 @@
...
+            for ir in setuptools_kwargs['install_requires']
+        ]
 
-    # rapidjson exists on Python 3 only
-    if sys.version_info >= (3, 0):
-        setuptools_kwargs['extras_require']['rapidjson'] = ['python-rapidjson']
-        setuptools_kwargs['tests_require'].append('python-rapidjson')
+    # 2020-03-23: for now, disable pyspark on Python 3.8 (it installs
+    # but doesn't work)
+    if sys.version_info[:2] >= (3, 4):
```
88manpreet: Should it be >= (3, 8)?

88manpreet:
+1

PR #2155, opened 2020-04-01
recursively merge extra_cluster_params

### LINKED ISSUE #2154: extra_cluster_params should merge dictionaries
Currently, if you do something like this in your config:

```
runners:
  emr:
    extra_cluster_params:
      Instances:
        EmrManagedMasterSecurityGroup: sg-00000000000000000
```

it'll actually blow away the entire `Instances` parameter to `RunJobFlow`, rather than just setting the security group. This is almost certainly not what users expect.

davidmarin:
Make something like this:

```
runners:
  emr:
    extra_cluster_params:
      Instances:
        EmrManagedMasterSecurityGroup: sg-00000000000000000
```

add `EmrManagedMasterSecurityGroup` to the `Instances` API parameter, rather than clobbering it.

Fixes #2154.

### DIFF
--- a/docs/guides/cloud-opts.rst
+++ b/docs/guides/cloud-opts.rst
@@ -364,6 +364,19 @@ Cluster software configuration
               SupportedProducts:
               - mapr-m3
 
+    .. versionchanged:: 0.7.2
+
+       Dictionaries will be recursively merged into existing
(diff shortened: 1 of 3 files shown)


88manpreet:
+1

PR #2156, opened 2020-04-08
default python_bin on Python 2 to 'python2'

### LINKED ISSUE #2151: default to 'python2.7', not 'python' when on Python 2
Python 2 has reached end-of-life and we can no longer assume it's the default Python installation. For example, `python` is Python 3 on EMR AMI 6.0.0

davidmarin:
Python 2 has reached end-of-life, so the default `python` binary is finally Python 3 on some systems (notably, EMR AMI 6.x). Fixes #2151.

`python_bin` on Python 3 still defaults to `python3`.

PyPy has a different naming convention, so the default binary for PyPy 2 is still `pypy`.


88manpreet:
+1

PR #2157, opened 2020-04-11
use 'python2.7' as default python_bin for Python 2

davidmarin:
It turns out that `python2` isn't as common a binary name as i though; this changes it to `python2.7`. This is a second try at fixing #2151, and supersedes pull request #2156.

This also allows us to remove a special case that was already defaulting python_bin to `python2.7` on some old AMIs.


88manpreet:
+1

Issue #2162, opened 2020-04-16
cluster locks are never released


davidmarin:
Currently, our pooling code will "lock" a cluster to make sure that two jobs won't get simultaneously added to the same pooled cluster.

However, there isn't really a mechanism for releasing a lock. The way it's worked until now is that the lock is effectively expired once a job adds its step(s), since the lock is based on both the cluster ID and the number of steps it has. Now that we're avoiding looking at pooled clusters' steps (see #2159), we need to release the lock explicitly. Basically, as soon as the cluster is no longer `WAITING`, it's safe to release the lock.

It also makes sense to expire cluster locks rather quickly. Basically, we need long enough for the user to submit steps and for the cluster to go into the `RUNNING` state (which might happen immediately; need to check this).

davidmarin:
Looks like it takes maybe 10 seconds for the cluster to notice the new steps and switch into the `RUNNING` state.

Probably best to have the lock automatically expire after a minute. Plus we can unlock it earlier if we see the cluster go into a non-`WAITING` state or on runner cleanup.

Issue #2159, opened 2020-04-15
don't list steps when pooling


davidmarin:
The pooling code uses API calls rather inefficiently. To find a cluster to join we:

 * use `ListClusters` to find all clusters in the `WAITING` state (in `_usable_clusters()`)
 * for each of these clusters, call `DescribeCluster` (also in `_usable_clusters()`, see `_check_cluster_state()` for the comparison)
 * list _all_ the cluster's steps (in `_check_cluster_state()`)
 * for each cluster that passes this step, get all its instance groups/fleets (in `_compare_cluster_setup()`)

There are several reasons we check a cluster's steps, and it turns out they're all outdated. First, we don't want to exceed the limit of 256 steps per cluster, but [that only applies](https://docs.aws.amazon.com/emr/latest/ManagementGuide/AddMoreThan256Steps.html) to clusters running AMIs prior to 3.1.1 (and 2.x clusters prior to 2.4.8). We could get around this by simply preventing the user from using pooling with such outdated AMIs.

Second, there is some defensive code (in `_check_cluster_state()`) that worried about clusters having steps but failing to run them. This was always a somewhat mysterious issue, but it's several years old, and we can assume it's been fixed.

Finally, we want to know which step number the cluster is on because we included that in the S3 URI we use to "lock" the cluster. The reason for this is obscure, and [goes back to the original locking code](https://github.com/Yelp/mrjob/commit/6f47a7ed70a0531ec799a033e1811d83501af7a3). Maybe the idea was to make locks self-expire once a step was added? In any case, there are many other ways to implement lock expiry, and they don't require knowledge of the step number.

davidmarin:
Before we do this, we need to have a mechanism for releasing cluster locks; see #2162.

PR #2163, opened 2020-04-18
stop calling ListSteps when using pooled clusters

davidmarin:
Currently, our pooling code makes the EMR API call `ListSteps` on every eligible cluster, which can be expensive in terms of API calls (see #2159, which this fixes, for details).

So that two runners don't both submit a job to a cluster at the same time, there is a "locking" mechanism which uses S3. The current code includes both the cluster ID and the number of steps in the cluster in the URI of the lock file, so that a lock will be implicitly released once steps are added to the cluster. Instead, runners now explicitly release locks once their step(s) start running (fixing #2162), so we can just use the cluster ID.

We also ignored the limit on total number of steps in very old AMIs, in favor of telling the user pooling might not work properly on them, and trusting that the race case where a cluster has pending steps but is still in the `WAITING` stage (which seems to last about 10 seconds) will be prevented by cluster locking.

We also now hard-coded cluster lock expiry to last one minute, since we're just trying to cover the time between step submission and when the cluster starts running the steps.

Before v0.7.3 is released, I'm planning to replace the S3 "locking" mechanism with EMR tags, so I put a little less effort into unit-testing the locking mechanism than I otherwise would. I did hand-test pooling, and it worked as expected.


mrjob/emr.py
```diff
@@ -920,6 +940,8 @@ def cleanup(self, mode=None):
             except Exception as e:
                 log.exception(e)
 
+        # TODO: otherwise, cancel any steps we submitted
```
88manpreet, line 943: Would it makes sense to add issue-id where it is supposed to be fixed?

mrjob/emr.py
```diff
@@ -2590,26 +2557,24 @@ def _find_cluster(self, num_steps=1):
...
-                    self.fs.s3, self._lock_uri(cluster_id, cluster_num_steps),
-                    self._opts['cloud_fs_sync_secs'], self._job_key)
-                if status:
-                    log.debug('Acquired lock on cluster %s', cluster_id)
+                    len(cluster_id_list),
+                    '' if len(cluster_id_list) == 1 else 's',
+                    ': ' if cluster_id_list else '',
+                    ', '.join(reversed(cluster_id_list))))
+
+            # TODO: why not just try locking every cluster in the list?
```
88manpreet, line 2570: Same here?

88manpreet:
+1

mrjob/emr.py
davidmarin: good point, done

mrjob/emr.py
davidmarin: good point, done

88manpreet:
+1

PR #2167, opened 2020-04-22
lock clusters with EMR tags

davidmarin:
This replaces the S3-based "locking" mechanism used by cluster pooling with one that uses EMR tags. This means that we no longer depend on all jobs in a pool using the same temp dir on S3, and we don't have to wait for S3 to sync. This also makes the timing of locking more consistent and less prone to race conditions. Fixes #2106 

`terminate_idle_clusters` no longer locks pooled clusters before shutting them down, both because I don't want to have to support both the old and new kind of locking forever, and because this script is just a safeguard; clusters terminate themselves now.


88manpreet:
+1, approved.

PR #2169, opened 2020-05-15
try all matching pooled clusters before asking for more

### LINKED ISSUE #2164: pooling should use any usable cluster before looking for more
For whatever reason, `_find_clusters()` will find a list of eligible clusters, but if the "best" cluster turns out not to be available because another job locked it, it doesn't try any of the others, instead calling `_usable_clusters()` repeatedly, which fires off more `DescribeCluster` and `ListInstance*` API calls.

This is extremely wasteful of API calls. We should try submitting our job to one of the other clusters in the list before trying again.

davidmarin:
_This builds on pull request #2167_

When pooling finds multiple matching clusters for the job to use, we now try joining all of them in turn, rather than giving up if we can't lock the "best" cluster. Fixes #2164 

This branch also changes pool locking semantics to only lock a cluster if it's in the `WAITING` state. This is necessary because if we attempt to lock a cluster and fail, our information about which clusters are available may no longer be up-to-date.


mrjob/tools/emr/terminate_idle_clusters.py
```diff
@@ -51,8 +51,7 @@
                         this higher than the amount of time your jobs can take
                         to start instances and bootstrap.
   --max-mins-locked MAX_MINS_LOCKED
-                        Max number of minutes a cluster can be locked while
-                        idle.
+                        Deprecated, does nothing
```
88manpreet, line 54: [Optional] Though, not really required but would it make sense to add [deprecation warning](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).

88manpreet:
+1

Issue #2176, opened 2020-05-27
Spark harness is not populating counters when counter-output-dir is not an S3 path


88manpreet:
[Hadoop Counters](https://data-flair.training/blogs/hadoop-counters/), an integral feature of Hadoop map-reduce provides a way to measure the progress or the number of operations that occur within map/reduce job.
[Spark harness runs](https://github.com/Yelp/mrjob/wiki/How-mrjob-works#the-spark-harness) the regular hadoop streaming job on spark using Spark Runner.
Spark harness emulates the [counters feature](https://mrjob.readthedocs.io/en/latest/guides/spark.html#counter-emulation-is-almost-perfect) to run the same hadoop-streaming job onto spark without requiring any modifications to the job.
Harness script stores the calculated counters value to the given counter_output_dir (--spark-tmp-dir) using [saveAsTextFile](https://github.com/Yelp/mrjob/blob/master/mrjob/spark/harness.py#L287) spark api. The populated counters values is in turn read by [spark runner](https://github.com/Yelp/mrjob/blob/master/mrjob/runner.py#L663) to be provided to application user.
The logic works well if the the path for `--spark-tmp-dir` is an S3 path. With the regular local file-path (by default), `saveAsTextfile` creates the counters file (part-*) on spark executors but not on the driver local file-path. Unless, the executors are running on the same host as drivers.

davidmarin:
Fixed by #2177.

PR #2177, opened 2020-05-27
Write counters to local filepath when spark-tmp-dir is not an S3 path.

88manpreet:
Issue: #2176 
The proposed fix is to write to given local-file path using python file writer if the part-* (counters file) is not created on driver.

Additional (possible?) benefit of the fix is to avoid additional communication overhead (latency and network access) of spark harness to populate S3 bucket and spark driver retrieving content from the S3 bucket besides additional operations like deleting the bucket. I noticed with my test-runs that with given s3 path the job was hanging for a long time.

Other changes include flake error fixes.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,7 @@ v0.7.3, 2020-05-?? -- cluster pooling 2.0
  * cluster pooling rules are slightly more strict
    * mrjob version must always match
    * application list must match exactly
+ * support for counters to be published with with counter-output-dir as a local file-path
 
 v0.7.2, 2020-04-11 -- archives on all Spark platforms
  * archives work on non-YARN Spark installations (#1993)
--- a/mrjob/spark/harness.py
+++ b/mrjob/spark/harness.py
@@ -17,9 +17,11 @@ import sys
 import json
 from argparse import ArgumentParser
 from collections import defaultdict
+from glob import glob
 from importlib import import_module
 from itertools import chain
 
+from mrjob.parse import is_uri
 from mrjob.util import shlex_split
 from pyspark.accumulators import AccumulatorParam
 
@@ -281,6 +283,8 @@ def main(cmd_line_args=None):
         if args.counter_output_dir is not None:
             counters = [ca.value for ca in counter_accumulators]
 
+            # If the given path is an s3 path, use s3.parallelize,
+            # otherwise just write them directly to the local dir
             sc.parallelize(
                 [json.dumps(counters)],
                 numSlices=1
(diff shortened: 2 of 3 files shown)


CHANGES.txt
```diff
@@ -1,10 +1,11 @@
-v0.7.3, 2020-05-?? -- cluster pooling 2.0
+v0.7.3, 2020-05-?? -- spark harness counters support improvement
  * terminate_idle_clusters no longer locks pooled clusters
  * cluster pooling rules are slightly more strict
    * mrjob version must always match
    * application list must match exactly
+ * support for counters to be published with with counter-output-dir as a local file-path (#2176)
 
-v0.7.2, 2020-04-11 -- archives on all Spark platforms
```
stug, line 7: Pretty sure this is wrong -- 0.7.2 was released this year.

mrjob/spark/harness.py
```diff
@@ -281,12 +284,22 @@ def make_mrc_job(mrc, step_num):
...
+            # otherwise just write them directly to the local dir
+            if is_uri(args.counter_output_dir):
+                sc.parallelize(
+                    [json.dumps(counters)],
+                    numSlices=1
+                ).saveAsTextFile(
+                    args.counter_output_dir
+                )
+                # Use regular python buildin file writer if the part-* file is not created
+            elif not glob(args.counter_output_dir + "/part-*"):
```
stug, line 297: I am not sure I understand why we need this check.  Are we just trying to avoid overwriting results from a previous run?  If so, that's now how we do it for the non-local counter_output_dir, so not sure why we'd need it here.  Or am I missing something?

tests/spark/test_harness.py
```diff
@@ -467,6 +478,32 @@ def test_increment_counter(self):
...
+                MRCountingJob, input_bytes=input_bytes,
+                counter_output_dir='{}'.format(output_counter_dir)
+            )
+            with harness_job.make_runner() as runner:
+                runner.run()
+                print(listdir(output_counter_dir))
+                path = join(output_counter_dir, "part-00000")
+                with open(path, 'r') as rb:
+                    counters = rb.read()
+                print('prince counters', counters)
```
stug, line 501: Are you leaving these `print()` calls here on purpose?

88manpreet: Nopes. Thanks for catching this.

mrjob/spark/harness.py
88manpreet: I will remove this check, this is not really required.

PR #2184, opened 2020-06-29
Docker support on EMR

### LINKED ISSUE #2179: support docker on EMR 6.x AMIs
EMR [now has explicit support for Docker on 6.x AMIs](https://aws.amazon.com/blogs/big-data/run-spark-applications-with-docker-using-amazon-emr-6-0-0-beta/). mrjob should integrate with this

davidmarin:
Adds support for running Spark executors in Docker containers on EMR. It's activated by setting the `docker_image` option. Also added the `docker_mounts` option, and the `docker_client_config` option (so you can log into ECR). Fixes #2179.

If you want to test this out without building a Docker image, try `python -m mrjob.examples.mr_spark_wordcount --docker-image amazoncorretto:8` on Python 2.


mrjob/cloud.py
```diff
@@ -77,9 +77,6 @@ class HadoopInTheCloudJobRunner(MRJobBinRunner):
         'zone',
     }
 
-    # so far, every service provides the ability to run bootstrap scripts
-    _BOOTSTRAP_MRJOB_IN_PY_FILES = False
```
88manpreet, line 81: Why is this removed?

mrjob/emr.py
```diff
@@ -712,17 +717,10 @@ def _check_output_not_exists(self):
     def _add_bootstrap_files_for_upload(self, persistent=False):
         """Add files needed by the bootstrap script to self._upload_mgr.
 
-        Tar up mrjob if bootstrap_mrjob is True.
-
         Create the master bootstrap script if necessary.
 
         persistent -- set by make_persistent_cluster()
         """
-        # lazily create mrjob.zip
```
88manpreet, line 721: Why are we removing this?

88manpreet:
Looks good to me. Left some questions. +1 for merging this.

mrjob/cloud.py
davidmarin: Because installing mrjob on the cluster at bootstrap time no longer helps when we're using Docker, since that won't impact what libraries are available inside the Docker container.

mrjob/emr.py
davidmarin: Because mrjob is no longer part of bootstrapping; it's now part of `py_files` like on the Hadoop runner.

PR #2188, opened 2020-07-23
Migrate from master and slave phrasing

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


davidmarin:
I'm aware of this issue, but I already removed "slave" from the code as much as I could, referring to "core" instances instead. I can't actually change the name of options in Hadoop, or error messages from the EMR API.

PR #2197, opened 2020-08-26
pooling: query YARN resource manager for available memory and CPU

### LINKED ISSUE #2191: join pooled clusters based on yarn cluster metrics
Currently, pooling will check if a cluster is "big enough" in terms of memory, CPU and a few other aspects (e.g. EBS volume size).

We can instead compare memory and CPU needs by SSHing to the cluster's YARN resource manager and querying its [metrics API](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Metrics_API) for `availableMB` and `availableVirtualCores`.

Not only would this provide more useful information about a cluster that can run multiple jobs simultaneously, it would also allow us to skip querying the cluster's instances `ListInstanceGroups`/`ListInstanceFleets`, saving an API call.

coyotemarin:
This change allows you to replace looking at a cluster's number and type of instances with connecting directly to the YARN resource manager and querying it for available memory and CPU. Fixes #2191.


88manpreet:
+1

PR #2203, opened 2020-09-16
v0.7.4 release

### LINKED ISSUE #2193: upgrade boto3/botocore to support StepConcurrencyLevel
Old versions of boto3/botocore don't support the `StepConcurrencyLevel` parameter to `RunJobFlow`. Need to find which version of boto3/botocore introduced support for it, and bump requirements accordingly.

### LINKED ISSUE #2196: useless return value from make_pooled_cluster() in pooling tests
`make_pooled_cluster()` in `tests/test_emr_pooling.py` returns a runner and a cluster ID, but the runner is never used. We should just return the cluster ID.

This is a very simple fix (and self-testing), but it should be a separate pull request, so as not to obscure more substantive changes.

coyotemarin:
Final documentation updates and style fixes for the v0.7.4 release.

This also updates boto3/botocore library versions (fixes #2193) and cleans up some tests of cluster pooling (fixes #2196).

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -4,6 +4,88 @@ What's New
 For a complete list of changes, see `CHANGES.txt
 <https://github.com/Yelp/mrjob/blob/master/CHANGES.txt>`_
 
+.. _v0.7.4:
+
+0.7.4
+-----
+
+Docker on EMR
+^^^^^^^^^^^^^
+
+This release adds support for Docker on EMR, which
+`was released with AMI version 6.0.0 <https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-spark-docker.html>`__.
+This is enabled by setting :mrjob-opt:`docker_image` to point at your image.
+
+There is also a :mrjob-opt:`docker_mounts` option, and, if you want to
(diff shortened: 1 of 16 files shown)


88manpreet (approved):

docs/whats-new.rst
```diff
@@ -4,6 +4,88 @@ What's New
...
+For this feature to work, you must enable SSH (the :mrjob-opt:`ec2_key_pair`
+and :mrjob-opt:`ec2_key_pair_file` options).
+
+You can now control the size of your cluster pool with the
+:mrjob-opt:`max_clusters_in_pool` option. If a job wants to launch a new
+cluster in the pool but the pool is already "full," it will wait and try again
+until the pool is no longer full or it can join a cluster.
+
+Once a job
+determines that it is okay to add another cluster to the pool, it will
```
88manpreet, line 64: Not an issue. Out of curiosity, is there a reason this is not a single line? I saw it at multiple places.

coyotemarin: My editor is set to warn about lines longer than 80 characters, so i usually break the lines myself.

PR #2208, opened 2021-03-31
Set Tags on EMR cluster creation

### LINKED ISSUE #2207: Assign tags on EMR creation in single API call
AWS has deprecated several managed EMR policies (see [here](https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-managed-iam-policies.html)) and the replacement policies make use of tags for authorization.  

Currently, tags are assigned *after* a cluster is made, in a [separate api call](https://github.com/Yelp/mrjob/blob/master/mrjob/emr.py#L1146), though the underlying boto3 client [supports setting tags on cluster creation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/emr.html#EMR.Client.run_job_flow).  

To avoid any issues that may arise with failed tag application - either due to a subsequent failure of the tags call, or because of the (albeit small) time between API calls - the tags should be assigned in the same API call as the cluster creation.

mgmarino:
This ensures that tags are set at the moment of cluster creation, which
can have implications for permissions requiring tags.

Closes #2207

For more information on this, see the documentation [here](https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-managed-iam-policies.html) and the portion about [tagging the EMR cluster](https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-managed-iam-policies.html#manually-tagged-resources)

### DIFF
--- a/mrjob/emr.py
+++ b/mrjob/emr.py
@@ -1121,12 +1121,20 @@ class EMRJobRunner(HadoopInTheCloudJobRunner, LogInterpretationMixin):
         emr_client = self.make_emr_client()
 
         kwargs = self._cluster_kwargs(persistent)
+        kwargs["Tags"] = self._build_tags()
         log.debug('Calling run_job_flow(%s)' % (
             ', '.join('%s=%r' % (k, v)
                       for k, v in sorted(kwargs.items()))))
         cluster_id = emr_client.run_job_flow(**kwargs)['JobFlowId']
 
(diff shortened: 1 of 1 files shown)


mgmarino:
Failing tests fixed in #2209

mgmarino:
Any input from the maintainers?

mgmarino:
Ping?

PR #2225, opened 2025-03-15
util.py (and hence mrjob) not to depend on pipes, removed in python 3.13

### LINKED ISSUE #2224: Python 3.13 support
I suppose https://github.com/Yelp/mrjob/issues/2222 has to be fixed first for 3.12 support, but after that we might also fix the support for 3.13.
In Python 3.13 the module called `pipes` has been removed, hence I got the following error:
`ModuleNotFoundError: No module named 'pipes'`
See: https://docs.python.org/3/library/pipes.html

dsevilla:
util.py (and hence mrjob) not to depend on pipes, module removed in python 3.13.

Fixes #2224.

### DIFF
--- a/mrjob/bin.py
+++ b/mrjob/bin.py
@@ -20,8 +20,8 @@
 import logging
 import os
 import os.path
-import pipes
 import re
+import shlex
 import sys
 from mrjob.py2 import PY2
 from platform import python_implementation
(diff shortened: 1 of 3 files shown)


mocusez:
image

Hi devilla，

I think fix `util.py` is not enough, the `bin.py` also use pipes as well

dsevilla:
You are right. I fixed it in my fork, but I think I didn't used the patch-1 branch so it does not get reflected here. I'll try to change the reference branch for the pull request...

dsevilla:
OK, I merged the changes in patch-1 now, as I cannot change the base branch for the pull request.

